PR #3, opened 2023-10-04
feat: migration to Alloy

Evalir:


test-data/test-contract-versions/pinned-0.4.14.sol
mattsse, line 1: what's with these tests?

mattsse (approved):
cool, this was simple, 

pending q re solidity test data

test-data/test-contract-versions/pinned-0.4.14.sol
Evalir: Noticed `alloy_json_abi::JsonAbi` doesn't parse pre-`0.5.x` ABIs and we had these test expecting this to be the case from ethabi.  This shouldn't be an issue for foundry as ds-test is `>=0.5.0` though, and we should prob just remove them00.

gakonst (approved):
let's cut the commented out tests and add a note about that in the readme -- lgtm

PR #19, opened 2023-11-07
more test in report/compiler.rs and Default trait for CompilerInput

DoTheBestToGetTheBest:


src/report/compiler.rs
```diff
@@ -208,4 +211,51 @@ mod tests {
         );
         std::env::remove_var("foundry_compilers_LOG");
     }
+
+    #[test]
+    fn fails_on_invalid_format() {
+        let result: Result<Target, _> = "invalid_format".parse();
+        assert!(result.is_err());
+    }
+    #[test]
```
Evalir, line 220: let's add some spacing in between the test functions

src/report/compiler.rs
```diff
@@ -208,4 +211,51 @@ mod tests {
...
+    fn fails_on_bad_name() {
+        let result: Result<Target, _> = "bad=in.json".parse();
```
Evalir, line 222: we can probably just get the err directly (as we know it's a bad input) with `.err().unwrap()`

src/report/compiler.rs
```diff
@@ -208,4 +211,51 @@ mod tests {
...
+        assert!(result.is_err());
+        if let Err(e) = result {
```
Evalir, line 224: see above, so that way we don't have conditional assertions

Evalir (changes requested):
more tests always better

pending nits & ci

Evalir:
ping @DoTheBestToGetTheBest

DoTheBestToGetTheBest:
> ping @DoTheBestToGetTheBest

Thank you for pinging me, going to check this !

DoTheBestToGetTheBest:
sorry for the late time, does this look good to you ? work fin in my terminal

src/artifacts/mod.rs
```diff
@@ -54,7 +54,7 @@ const SOLIDITY: &str = "Solidity";
 const YUL: &str = "Yul";
 
 /// Input type `solc` expects
-#[derive(Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize, Default)]
```
mattsse, line 57: we should implement default manually and set the language to `SOLIDITY.to_string()`

mattsse (changes requested):

src/artifacts/mod.rs
DoTheBestToGetTheBest: > we should implement default manually and set the language to `SOLIDITY.to_string()`

Thank you Mat, does this looks better?

src/report/compiler.rs
```diff
@@ -208,4 +211,53 @@ mod tests {
...
+    }
+
+    #[test]
+    fn fails_on_bad_name() {
+        let result: Result<Target, _> = "bad=in.json".parse();
+        assert!(result.is_err());
+        if let Err(e) = result {
+            assert_eq!(e.to_string(), "bad=in.json");
+        }
+    }
```
mattsse, line 228: these aren't that useful, please remove

mattsse (changes requested):
nit

src/report/compiler.rs
DoTheBestToGetTheBest: > these aren't that useful, please remove

done, ty

mattsse (approved):

Evalir (approved):
Thank you!

DoTheBestToGetTheBest:
ty  @mattsse and @Evalir

PR #29, opened 2023-11-15
 Log Management with File Rotation in SolcCompilerIoReporter

DoTheBestToGetTheBest:
this pr introduce a log file rotation to SolcCompilerInReport, it is supposed to handle large log file generated during compilation processes

This should prevents log fles from growing too much large.


DaniPopes:
We're planning to refactor this code, closing for now as stale.

PR #35, opened 2023-12-24
fix: create valid Standard JSON to verify for projects with symlinks

tash-2s:
Currently, when a project includes symbolic links to Solidity files, `Project::standard_json_input` returns a Standard JSON Input struct that is unusable for contract verification on block explorers. `solc` cannot compile these contracts based on the json, leading to verification failures. This pull request addresses this issue.

I encountered this problem when using the `forge verify-contract` command, which internally uses this function. While `forge build` completes successfully, the verification command fails, indicating that the block explorer cannot find a source file.

This problem occurs in projects that contain symlinked Solidity files and when these files are imported from other files. My project, structured as a monorepo, uses external libraries installed via pnpm. These libraries, accessible from the `./node_modules/` directory, are set up with remappings in my Foundry project. However, directories under `./node_modules/` are symlinks pointing to other locations, leading to this issue.

## Reproduction

I added a test named `can_create_standard_json_input_with_symlink` to demonstrate the issue within this repository. Also, the error can be reproduced using the `forge verify-contract` command and steps below:

Environment: macOS (Apple silicon), forge 0.2.0 (`c312c0d` 2023-12-22T00:20:29.297186000Z)

```console
$ mkdir repro && cd repro

$ mkdir -p project/src project/node_modules dependency

$ cat << EOF > project/src/A.sol
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.23;
import "@dependency/B.sol";
contract A is B {}
EOF

$ cat << EOF > dependency/B.sol
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.23;
import "./C.sol";
contract B is C {}
EOF

$ cat << EOF > dependency/C.sol
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.23;
contract C {}
EOF

$ cat << EOF > project/foundry.toml
[profile.default]
remappings = ["@dependency/=node_modules/dependency/"]
allow_paths = ["../dependency/"]
EOF

# Create a symbolic link
$ cd project/node_modules
$ ln -s ../../dependency dependency
$ cd ../..

# Display the file structure
$ tree
.
├── dependency
│   ├── B.sol
│   └── C.sol
└── project
    ├── foundry.toml
    ├── node_modules
    │   └── dependency -> ../../dependency
    └── src
        └── A.sol

# `build` succeeds
$ cd project
$ forge build

# `verify-contract` generates an unintended json (C.sol has a host absolute path name)
$ forge verify-contract --show-standard-json-input 0x0000000000000000000000000000000000000000 A | jq . > A.json
$ cat A.json
{
  "language": "Solidity",
  "sources": {
    "src/A.sol": {
      "content": "// SPDX-License-Identifier: UNLICENSED\npragma solidity ^0.8.23;\nimport \"@dependency/B.sol\";\ncontract A is B {}\n"
    },
    "node_modules/dependency/B.sol": {
      "content": "// SPDX-License-Identifier: UNLICENSED\npragma solidity ^0.8.23;\nimport \"./C.sol\";\ncontract B is C {}\n"
    },
    "/Users/t/bench/repro/dependency/C.sol": {
      "content": "// SPDX-License-Identifier: UNLICENSED\npragma solidity ^0.8.23;\ncontract C {}\n"
    }
  },
  "settings": {
    "remappings": [
      "@dependency/=node_modules/dependency/",
      "dependency/=node_modules/dependency/"
    ],
    "optimizer": {
      "enabled": true,
      "runs": 200
    },
    "metadata": {
      "useLiteralContent": false,
      "bytecodeHash": "ipfs",
      "appendCBOR": true
    },
    "outputSelection": {
      "*": {
        "": [
          "ast"
        ],
        "*": [
          "abi",
          "evm.bytecode",
          "evm.deployedBytecode",
          "evm.methodIdentifiers",
          "metadata"
        ]
      }
    },
    "evmVersion": "paris",
    "libraries": {}
  }
}

# `solc` cannot compile using the json
$ solc --standard-json --no-import-callback A.json | jq .errors
[
  {
    "component": "general",
    "errorCode": "6275",
    "formattedMessage": "ParserError: Source \"node_modules/dependency/C.sol\" not found: No import callback.\n --> node_modules/dependency/B.sol:3:1:\n  |\n3 | import \"./C.sol\";\n  | ^^^^^^^^^^^^^^^^^\n\n",
    "message": "Source \"node_modules/dependency/C.sol\" not found: No import callback.",
    "severity": "error",
    "sourceLocation": {
      "end": 81,
      "file": "node_modules/dependency/B.sol",
      "start": 64
    },
    "type": "ParserError"
  }
]
```

Manually editing the host filesystem's absolute path as shown below allows the `solc` command to succeed.

```diff
   "node_modules/dependency/B.sol": {
     "content": "// SPDX-License-Identifier: UNLICENSED\npragma solidity ^0.8.23;\nimport \"./C.sol\";\ncontract B is C {}\n"
   },
-  "/Users/t/bench/repro/dependency/C.sol": {
+  "node_modules/dependency/C.sol": {
     "content": "// SPDX-License-Identifier: UNLICENSED\npragma solidity ^0.8.23;\ncontract C {}\n"
   }
 },
```

## Cause

The issue arises because import paths in Solidity files are processed by the [`std::fs::canonicalize`](https://doc.rust-lang.org/std/fs/fn.canonicalize.html) function. This function resolves symlinks and normalizes paths. When symlinks are resolved, it results in `solc` being unable to locate the corresponding source paths in the json, as it relies on Solidity import statements. Therefore, symlinks should not be resolved here. The paths should be maintained as specified in the Solidity files, except for basic normalization.

## Solution

To address this, I implemented an import path normalization function and replaced the canonicalization function where necessary. Based on [the Solidity documentation page](https://docs.soliditylang.org/en/v0.8.23/path-resolution.html), this function resolves `.` and `..` segments/components without resolving symlinks.

The Standard JSON's source paths, for verification purposes, should be based on the project root. This allows the compiler to find sources within the json. The conversion of normalized paths to project root-based paths occurs after all sources are processed. That conversion is already implemented, so this PR doesn't need to address it. ([It seems like the path conversion needs improvement](https://github.com/foundry-rs/foundry/issues/5307), but it is a separate issue and should be handled in another PR.)

https://github.com/foundry-rs/compilers/blob/b1561d807c246c066c7c4b0c72c8eb64c696a43d/src/lib.rs#L514-L525

With the changes proposed in this PR, I confirmed the fix of the issue by building `forge` with this patch and testing both the reproduction case and my project. The resulting json matches the manually edited json diff mentioned earlier.

A potential downside of this approach is that the same file could be represented differently in the json if accessed via both symlinked and real paths. However, I see no issues with this, and it aligns with the behavior of the compiler's virtual filesystem.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -200,6 +200,35 @@ pub fn normalize_solidity_import_path(
     fs::metadata(&normalized).map(|_| normalized).map_err(|err| SolcIoError::new(err, original))
 }
 
+// This function lexically cleans the given path.
+//
+// It performs the following transformations for the path:
+//
+// * Resolves references (current directories (`.`) and parent (`..`) directories).
+// * Reduces repeated separators to a single separator (e.g., from `//` to `/`).
+//
+// This transformation is lexical, not involving the file system, which means it does not account
+// for symlinks. This approach has a caveat. For example, consider a filesystem-accessible path
+// `a/b/../c.sol` passed to this function. It returns `a/c.sol`. However, if `b` is a symlink,
+// `a/c.sol` might not be accessible in the filesystem in some environments. Despite this, it's
+// unlikely that this will pose a problem for our intended use.
+//
+// # How it works
+//
+// The function splits the given path into components, where each component roughly corresponds to a
+// string between separators. It then iterates over these components (starting from the leftmost
+// part of the path) to reconstruct the path. The following steps are applied to each component:
+//
+// * If the component is a current directory, it's removed.
+// * If the component is a parent directory, the following rules are applied:
+//     * If the preceding component is a normal, then both the preceding normal component and the
+//       parent directory component are removed. (Examples of normal components include `a` and `b`
+//       in `a/b`.)
+//     * Otherwise (if there is no preceding component, or if the preceding component is a parent,
+//       root, or prefix), it remains untouched.
+// * Otherwise, the component remains untouched.
+//
+// Finally, the processed components are reassembled into a path.
 fn clean_solidity_path(original_path: impl AsRef<Path>) -> PathBuf {
     let mut new_path = Vec::new();
 
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -177,6 +177,52 @@ pub fn canonicalize(path: impl AsRef<Path>) -> Result<PathBuf, SolcIoError> {
     res.map_err(|err| SolcIoError::new(err, path))
 }
 
+/// Returns a normalized Solidity file path for the given import path based on the specified
+/// directory.
+///
+/// This function resolves `./` and `../`, but, unlike [`canonicalize`], it does not resolve
+/// symbolic links.
+///
+/// The function returns an error if the normalized path does not exist in the file system.
+///
+/// See also: <https://docs.soliditylang.org/en/v0.8.23/path-resolution.html>
+pub fn normalize_solidity_import_path(
+    directory: impl AsRef<Path>,
+    import_path: impl AsRef<Path>,
+) -> Result<PathBuf, SolcIoError> {
+    let original = directory.as_ref().join(import_path);
+    let cleaned = clean_solidity_path(&original);
+
+    // this is to align the behavior with `canonicalize`
+    use path_slash::PathExt;
+    let normalized = PathBuf::from(dunce::simplified(&cleaned).to_slash_lossy().as_ref());
+
+    fs::metadata(&normalized).map(|_| normalized).map_err(|err| SolcIoError::new(err, original))
+}
+
+fn clean_solidity_path(original_path: impl AsRef<Path>) -> PathBuf {
+    let mut new_path = Vec::new();
+
+    for component in original_path.as_ref().components() {
+        match component {
+            Component::Prefix(..) | Component::RootDir | Component::Normal(..) => {
+                new_path.push(component);
+            }
+            Component::CurDir => {}
+            Component::ParentDir => match new_path.last() {
+                Some(Component::Normal(..)) => {
+                    new_path.pop();
+                }
+                _ => {
+                    new_path.push(component);
+                }
+            },
+        }
+    }
+
+    new_path.iter().collect()
+}
+
 /// Returns the same path config but with canonicalized paths.
 ///
 /// This will take care of potential symbolic linked directories.
@@ -228,7 +274,7 @@ pub fn resolve_library(libs: &[impl AsRef<Path>], source: impl AsRef<Path>) -> O
 /// until the `root` is reached.
 ///
 /// If an existing file under `root` is found, this returns the path up to the `import` path and the
-/// canonicalized `import` path itself:
+/// normalized `import` path itself:
 ///
 /// For example for following layout:
 ///
@@ -247,7 +293,7 @@ pub fn resolve_absolute_library(
 ) -> Option<(PathBuf, PathBuf)> {
     let mut parent = cwd.parent()?;
     while parent != root {
-        if let Ok(import) = canonicalize(parent.join(import)) {
+        if let Ok(import) = normalize_solidity_import_path(parent, import) {
             return Some((parent.to_path_buf(), import));
         }
         parent = parent.parent()?;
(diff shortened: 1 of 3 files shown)


src/utils.rs
```diff
@@ -177,6 +177,52 @@ pub fn canonicalize(path: impl AsRef<Path>) -> Result<PathBuf, SolcIoError> {
...
+    let cleaned = clean_solidity_path(&original);
+
+    // this is to align the behavior with `canonicalize`
+    use path_slash::PathExt;
+    let normalized = PathBuf::from(dunce::simplified(&cleaned).to_slash_lossy().as_ref());
+
+    fs::metadata(&normalized).map(|_| normalized).map_err(|err| SolcIoError::new(err, original))
+}
+
+fn clean_solidity_path(original_path: impl AsRef<Path>) -> PathBuf {
```
tash-2s, line 203: Rust doesn't have this normalization function (resolves `.` and `..` but doesn't resolve symlinks), so I wrote the function myself. The function is simple, but I found several crates that perform similar tasks, so I can switch to one of them if preferred.
e.g.
- https://crates.io/crates/path-clean
- https://crates.io/crates/path-absolutize
- https://crates.io/crates/lexiclean

Evalir, line 203: Could we add some docs here? I know it's an internal function, but having docs is nice and useful to revisit this later whenever needed

Evalir: gotcha—I'll defer to @mattsse here, but the function is so small I don't mind keeping it here instead of pulling in an additional dep, as we also have tests.

Evalir (commented):
this is awesome, thank you _so_ much!

I haven't reviewed this deeply yet, just some nits:

tash-2s:
Thank you for your review! I've added the doc to the function.

mattsse (commented):
this is awesome, ty!

symlinks man...

before we merge this, could you please open a foundry PR with the foundry-compilers crate patched to this PR to see if there are any sideffects?

tash-2s:
@mattsse Thank you for taking the time to review this pull request.

I've created [a companion PR in foundry](https://github.com/foundry-rs/foundry/pull/6675) with this patch. Also, I have confirmed that the tests in foundry passed successfully.

src/utils.rs
```diff
@@ -177,6 +177,81 @@ pub fn canonicalize(path: impl AsRef<Path>) -> Result<PathBuf, SolcIoError> {
...
+    import_path: impl AsRef<Path>,
+) -> Result<PathBuf, SolcIoError> {
+    let original = directory.as_ref().join(import_path);
+    let cleaned = clean_solidity_path(&original);
+
+    // this is to align the behavior with `canonicalize`
+    use path_slash::PathExt;
+    let normalized = PathBuf::from(dunce::simplified(&cleaned).to_slash_lossy().as_ref());
+
+    fs::metadata(&normalized).map(|_| normalized).map_err(|err| SolcIoError::new(err, original))
```
mattsse, line 200: what's this call for?

src/utils.rs
```diff
@@ -177,6 +177,81 @@ pub fn canonicalize(path: impl AsRef<Path>) -> Result<PathBuf, SolcIoError> {
...
+                new_path.push(component);
+            }
+            Component::CurDir => {}
+            Component::ParentDir => match new_path.last() {
+                Some(Component::Normal(..)) => {
+                    new_path.pop();
+                }
+                _ => {
+                    new_path.push(component);
+                }
```
mattsse, line 247: I think we can use `if let Some(Component::Normal) = new_path.last() {pop} else {...}` here

which is a bit nicer imo

mattsse (changes requested):
a few nits

src/utils.rs
tash-2s: The function `normalize_solidity_import_path` checks if the returned path exists in the file system. The use of `fs::metadata(&normalized)` is to obtain an `io::Error` for non-existing files without reading their entire contents. To return the same type of error as `utils::canonicalize`, an `io::Error` is necessary for `SolcIoError`.

While there is a specific function, `Path::is_file()`, that confirms if a path exists in the file system, it only returns a bool.

https://doc.rust-lang.org/std/path/struct.Path.html#method.is_file

> This is a convenience function that coerces errors to false. If you want to check errors, call `fs::metadata` and handle its `Result`.

However, I have now realized that there's an alias `Path::metadata()` for paths. Therefore, I will switch to using this function and add a comment.

https://doc.rust-lang.org/std/path/struct.Path.html#method.metadata

tash-2s:
I've addressed these comments.

mattsse (approved):
again, tysm for these fixes! 🙏

tash-2s:
Appreciate your review on the PR. 🙏

PR #36, opened 2023-12-26
fix: create verifiable Standard JSON for projects with external files

tash-2s:
- Resolves https://github.com/foundry-rs/foundry/issues/5307

Currently, Foundry projects containing Solidity files outside the project root directory face contract verification failures on block explorers. This issue occurs from the Standard JSON including unusable source paths for external files, represented as full absolute paths in their host file systems.

This PR addresses the issue by improving the path conversion process. For files not located under the project root directory, relative parent directory paths (`..`) are used, enabling the compiler to find the files within the json.

Steps to reproduce the issue are detailed in the linked issue above. Additionally, a test case representing that scenario has been added.

With this change, the json created in the reproduction scenario will appear as follows:

```json
{
  "language": "Solidity",
  "sources": {
    "src/Counter.sol": {
      "content": "// SPDX-License-Identifier: UNLICENSED\npragma solidity ^0.8.13;\n\nimport \"@remapped/Parent.sol\";\n\ncontract Counter {\n    uint256 public number;\n\n    function setNumber(uint256 newNumber) public {\n        number = newNumber;\n    }\n\n    function increment() public {\n        number++;\n    }\n}\n"
    },
    "../remapped/Parent.sol": {
      "content": "// SPDX-License-Identifier: UNLICENSED\npragma solidity ^0.8.13;\nimport \"./Child.sol\";\n"
    },
    "../remapped/Child.sol": {
      "content": "// SPDX-License-Identifier: UNLICENSED\npragma solidity ^0.8.13;\n"
    }
  },
  "settings": {
    "remappings": [
      "@remapped/=../remapped/",
      "ds-test/=lib/forge-std/lib/ds-test/src/",
      "forge-std/=lib/forge-std/src/"
    ],
    "optimizer": {
      "enabled": true,
      "runs": 200
    },
    "metadata": {
      "useLiteralContent": false,
      "bytecodeHash": "ipfs",
      "appendCBOR": true
    },
    "outputSelection": {
      "*": {
        "": [
          "ast"
        ],
        "*": [
          "abi",
          "evm.bytecode",
          "evm.deployedBytecode",
          "evm.methodIdentifiers",
          "metadata"
        ]
      }
    },
    "evmVersion": "paris",
    "libraries": {}
  }
}
```

The source path is now aligned with the project root.

I have successfully deployed and verified the contract on Etherscan using this change.

`forge create --rpc-url "wss://ethereum-holesky.publicnode.com" --verify --verifier-url "https://api-holesky.etherscan.io/api" --etherscan-api-key "..." --private-key "..." src/Counter.sol:Counter`

https://holesky.etherscan.io/address/0xe08c332706185521fc8bc2b224f67adf814b1880#code

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -947,15 +947,31 @@ impl<T: ArtifactOutput> ArtifactOutput for Project<T> {
 
 // Rebases the given path to the base directory lexically.
 //
-// The returned path from this function usually starts either with a normal component (e.g., `src`)
-// or a parent directory component (i.e., `..`), which is based on the base directory. Additionally,
-// this function converts the path into a UTF-8 string and replaces all separators with forward
-// slashes (`/`).
+// For instance, given the base `/home/user/project` and the path `/home/user/project/src/A.sol`,
+// this function returns `src/A.sol`.
 //
-// The rebasing process is as follows:
+// This function transforms a path into a form that is relative to the base directory. The returned
+// path starts either with a normal component (e.g., `src`) or a parent directory component (i.e.,
+// `..`). It also converts the path into a UTF-8 string and replaces all separators with forward
+// slashes (`/`), if they're not.
 //
-// 1. Remove the leading components from the path that match the base components.
-// 2. Prepend `..` components to the path, equal in number to the remaining base components.
+// The rebasing process can be conceptualized as follows:
+//
+// 1. Remove the leading components from the path that match those in the base.
+// 2. Prepend `..` components to the path, matching the number of remaining components in the base.
+//
+// # Examples
+//
+// `rebase_path("/home/user/project", "/home/user/project/src/A.sol")` returns `src/A.sol`. The
+// common part, `/home/user/project`, is removed from the path.
+//
+// `rebase_path("/home/user/project", "/home/user/A.sol")` returns `../A.sol`. First, the common
+// part, `/home/user`, is removed, leaving `A.sol`. Next, as `project` remains in the base, `..` is
+// prepended to the path.
+//
+// On Windows, paths like `a\b\c` are converted to `a/b/c`.
+//
+// For more examples, see the test.
 fn rebase_path(base: impl AsRef<Path>, path: impl AsRef<Path>) -> PathBuf {
     use path_slash::PathExt;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -518,7 +518,26 @@ impl<T: ArtifactOutput> Project<T> {
                 let path: PathBuf = if let Ok(stripped) = path.strip_prefix(root) {
                     stripped.to_slash_lossy().into_owned().into()
                 } else {
-                    path.to_slash_lossy().into_owned().into()
+                    let mut new_path = path.components().collect::<std::collections::VecDeque<_>>();
+
+                    for (i, (root_component, path_component)) in
+                        root.components().zip(path.components()).enumerate()
+                    {
+                        if root_component == path_component {
+                            new_path.pop_front();
+                        } else {
+                            let mut parent_dirs = vec![
+                                std::path::Component::ParentDir;
+                                root.components().collect::<Vec<_>>().len()
+                                    - i
+                            ];
+                            parent_dirs.extend(new_path);
+                            new_path = parent_dirs.into();
+                            break;
+                        }
+                    }
+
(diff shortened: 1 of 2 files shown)


tests/project.rs
```diff
@@ -1600,6 +1600,71 @@ fn can_sanitize_bytecode_hash() {
...
+    // ├── verif
+    // │   └── src
+    // │       └── Counter.sol
+    // └── remapped
+    //     ├── Child.sol
+    //     └── Parent.sol
+
+    let dir = tempfile::tempdir().unwrap();
+    let verif_dir = utils::canonicalize(dir.path()).unwrap().join("verif");
+    let remapped_dir = utils::canonicalize(dir.path()).unwrap().join("remapped");
```
tash-2s, line 1617: After [my previous PR](https://github.com/foundry-rs/compilers/pull/35) is merged, this canonicalization can be removed. It is necessary on this branch because the symlink-resolving canonicalization is still in effect, and the temporary directory on macOS is a symlink.

mattsse (commented):
amazing, tysm!

as mentioned in the other PR, I'd like to see a foundry companion PR with patched crate dep as a sanity check before we merge this

tash-2s:
Thank you for your review!

I've created [a companion PR in foundry](https://github.com/foundry-rs/foundry/pull/6676) with this patch. Also, I have confirmed that the tests in foundry passed successfully.

tash-2s:
The workflow test for this PR failed due to a failure in nextest installation. As this issue is unrelated to the changes made, re-running the test should resolve it. https://github.com/foundry-rs/compilers/actions/runs/7331166894/job/19979490471

src/lib.rs
```diff
@@ -954,6 +945,43 @@ impl<T: ArtifactOutput> ArtifactOutput for Project<T> {
...
+            new_path.pop_front();
+        } else {
+            let mut parent_dirs =
+                vec![std::path::Component::ParentDir; base.components().count() - i];
+            parent_dirs.extend(new_path);
+            new_path = parent_dirs.into();
+
+            break;
+        }
+    }
```
mattsse, line 980: I think we can solve this a bit differently
don't zip the two iterators but create the two iters separately and then iterate over the `path`
and continue as long as base and path components match

this way we don't need the vecdeque

mattsse (changes requested):

src/lib.rs
tash-2s: Hmm, I'm not sure I understand your suggestion.

If we compare the components of base and path at the same positions, it seems similar to using zip.

Could you please elaborate on the type of implementation you have in mind?

mattsse: something like

```rust
let mut components = Vec::new(); // perhaps we can even directly push components to a PathBuf?
let mut base_iter = base.components();
let mut path_iter = path.components();

while let Some(path) = path_iter.next() {
  if Some(path) != path_iter.next() {
     components.push(path);
     components.extend(path_iter);
  }
}


```

tash-2s: Thank you for your explanation. I now understand your point. I've made the change and pushed it.

src/lib.rs
```diff
@@ -954,6 +945,46 @@ impl<T: ArtifactOutput> ArtifactOutput for Project<T> {
...
+//
+// The returned path from this function usually starts either with a normal component (e.g., `src`)
+// or a parent directory component (i.e., `..`), which is based on the base directory. Additionally,
+// this function converts the path into a UTF-8 string and replaces all separators with forward
+// slashes (`/`).
+//
+// The rebasing process is as follows:
+//
+// 1. Remove the leading components from the path that match the base components.
+// 2. Prepend `..` components to the path, equal in number to the remaining base components.
```
mattsse, line 958: can we add an example here that illustrates this, otherwise this is a bit hard to understand.

mattsse (changes requested):
last pedantic doc nit

src/lib.rs
tash-2s: I've updated the comment, adding examples to make it easier to understand.

mattsse (approved):
awesome job!

I'll merge both and upstream to foundry after a new release

PaulRBerg:
Amazing work.

Doesn't this PR also address issue https://github.com/foundry-rs/foundry/issues/3507?

PR #39, opened 2024-01-04
feat: use Box<dyn> in sparse functions

DaniPopes:
To avoid boxing twice in case that an argument is already a box.


Evalir (approved):
makes sense!

PR #40, opened 2024-01-04
fix: account for unicode width in error syntax highlighting

DaniPopes:


src/artifacts/mod.rs
```diff
@@ -9,12 +9,14 @@ use serde::{de::Visitor, Deserialize, Deserializer, Serialize, Serializer};
...
-    ops::Range,
     path::{Path, PathBuf},
     str::FromStr,
     sync::Arc,
 };
-use yansi::{Color, Paint, Style};
+
+pub mod error;
+#[allow(ambiguous_glob_reexports)]
+pub use error::*;
```
DaniPopes, line 19: There's 3 different `SourceUnit` in this module.

mattsse: hmm, I was only able to find

https://github.com/foundry-rs/compilers/blob/38afb36147b74f8c3918431f5aa5fb7e8e15d1f7/src/artifacts/ast/mod.rs#L38-L38

DaniPopes: Sorry, SourceLocation

mattsse (approved):

PR #45, opened 2024-01-19
Rewrite dirty files discovery

klkvr:
## Motivation

Currently, dirty files discovery is not going deep enough into the graph, thus it often fails to discover all files that have dirty imports of some depth > 1. In worst case we may only recompile directly changed files and files directly importing them.

This PR changes this, so all files that have dirty files at some depth of their imports are marked as dirty.

This should solve a lot of weird issues with verification because it is very simple to occur such bug and it can result in, for example, script having and deploying outdated bytecode of contract which does not match the actual artifact.

## Solution

If we would consider that problem as a graph theory problem, than we are dealing with oriented graph of `N` nodes representing files and `M` oriented edges between them representing imports. Some of `N` nodes are marked as dirty initially and we want to mark as dirty all nodes from which a path exists to any of the dirty nodes.

Before the fix, the incorrect solution complexity was around `O(M)`, so I didn't want to increase it too much.

One possible solution without much changes to code could be to run deeper recursion from each node and try to reach dirty node from it. However, such solution would have a complexity of `O(N*M)` in worst case.

The solution I've implemented is working by precomputing `rev_edges` to store reversed edges, thus simplifying the problem to just finding all nodes that are reachable from the dirty ones in the reversed graph. The solution works by running DFS from all nodes that are known to be dirty in the beginning and keeping track of visited (dirty) nodes, to avoid visiting the same node twice. The complexity of such approach is `O(M)` in the worst case because we never use the same edge twice.

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -683,9 +683,10 @@ impl<'a, T: ArtifactOutput> ArtifactsCacheInner<'a, T> {
         // separates all source files that fit the criteria (dirty) from those that don't (clean)
         let mut dirty_sources = BTreeMap::new();
         let mut clean_sources = Vec::with_capacity(sources.len());
-        let mut memo = HashMap::with_capacity(sources.len());
+        let dirty_files = self.get_dirty_files(&sources, version);
+
         for (file, source) in sources {
-            let source = self.filter_source(file, source, version, &mut memo);
+            let source = self.filter_source(file, source, &dirty_files);
             if source.dirty {
                 // mark all files that are imported by a dirty file
                 imports_of_dirty.extend(self.edges.all_imported_nodes(source.idx));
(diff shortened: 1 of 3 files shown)


tests/project.rs
```diff
@@ -2731,3 +2731,55 @@ contract ContractTest {
     compiled.assert_success();
     assert!(compiled.find_first("Contract").is_some());
 }
+
+// This is a repro and a regression test for https://github.com/foundry-rs/compilers/pull/45
+#[test]
+fn dirty_files_discovery() {
```
Evalir, line 2737: Wonder if we can have another test or two, that explicitly check for proper recompilation of only dirty files? And maybe also a test like this, but it's all libraries but the top contract.

src/cache.rs
```diff
@@ -717,39 +718,42 @@ impl<'a, T: ArtifactOutput> ArtifactsCacheInner<'a, T> {
...
-        file: &Path,
-        version: &Version,
-        memo: &mut HashMap<PathBuf, bool>,
-        check_imports: bool,
-    ) -> bool {
-        match memo.get(file) {
-            Some(&dirty) => {
-                trace!(dirty, "memoized");
-                dirty
+    /// Returns a set of files that are dirty itself or import dirty file directly or indirectly
```
Evalir, line 728:
```suggestion
    /// Returns a set of files that are dirty itself or import dirty file directly or indirectly.
```

Evalir (commented):
This seems correct to me, after a thorough read. Love the approach and solution, and it does seem accurate—we should be marking as dirty all files which import, either directly or indirectly, a dirty file, and I think we've seen this out in the wild, with cleaning the cache being the solution.

Just some small nits, but I'll leave approval & merging to @mattsse

src/resolver/mod.rs
```diff
@@ -72,6 +72,8 @@ pub struct GraphEdges {
     /// The indices of `edges` correspond to the `nodes`. That is, `edges[0]`
     /// is the set of outgoing edges for `nodes[0]`.
     edges: Vec<Vec<usize>>,
+    /// Reverse of `edges`. That is, `rev_edges[0]` is the set of incoming edges for `nodes[0]`.
+    rev_edges: Vec<Vec<usize>>,
```
mattsse, line 76: ah, that makes sense

src/cache.rs
```diff
@@ -717,39 +718,42 @@ impl<'a, T: ArtifactOutput> ArtifactsCacheInner<'a, T> {
...
+    /// Accepts known dirty file and performs DFS over it's importers marking all visited files as
+    /// dirty.
+    #[instrument(level = "trace", skip_all, fields(file = %file.display()))]
+    fn populate_dirty_files(&self, file: &Path, dirty_files: &mut HashSet<PathBuf>) {
+        for file in self.edges.importers(file) {
+            // If file is marked as dirty we either have already visited it or it was marked as
+            // dirty initially and will be visited at some point later.
+            if !dirty_files.contains(file) {
+                dirty_files.insert(file.to_path_buf());
+                self.populate_dirty_files(file, dirty_files);
```
mattsse, line 756: I don't think this recursion will cause stack issues, but I'd rather be more cautious here and solve this iteratively if possible.

I guess this could be solved with a `stack: VecDeque` or similar

mattsse (changes requested):
this is great!

makes sense!
left a suggestion re recursion

src/cache.rs
Evalir: The previous solution was also recursive fwiw, but I do prefer an iterative approach to reduce any stack usage and so we can fail more gracefully in the weird case something happens during the DFS.

mattsse (approved):
I see the previous solution was also recursion without any issues, so this should be fine

wdyt @klkvr ?

Evalir (approved):
this lgtm. wdyt about a foundry companion PR to run CI over this and check with other repos @mattsse / @klkvr ?

klkvr:
previously recursion couldn't approach depth more than 2

the limit of recursion depth is 128, right? this can only be reached only if there is an import path with 128 files in a project, which shouldn't happen for most of the codebases

PR #50, opened 2024-01-24
Small fixes to typed AST

klkvr:
Add `Hash` derive for `SourceLocation`, fix schema for `TryCatchClause` and add test for it.

I am currently working on https://github.com/foundry-rs/foundry/issues/4034 and it requires usage of native solc AST, so I might be opening some other PRs around it as it seems that we are not using current models anywhere and they might be a bit outdated


mattsse (approved):
lgtm!

Issue #46, opened 2024-01-20
feat: `Cancun` EVM version


clabby:
## Overview

`solc` recently added `Cancun` as an EVM version in `EVMVersion.h` [here](https://github.com/ethereum/solidity/blob/develop/liblangutil/EVMVersion.h#L62). `EVMVersions`' doc comment [calls for syncing up](https://github.com/foundry-rs/compilers/blob/6528e4ac0b8f17599fd4b6b1e091b21c690ad142/src/artifacts/mod.rs#L781), which will unblock further support in `foundry`.

mattsse:
ref https://github.com/foundry-rs/compilers/blob/a420980485681d303c80519eb4dbedbe0caee899/src/artifacts/mod.rs#L783

PR #51, opened 2024-01-26
feat: add EVM version Cancun

DaniPopes:
I believe this is all one has to do.
Note that this does not support `0.8.24` as a Solc version anywhere else other than the `EvmVersion` enum.
Closes https://github.com/foundry-rs/compilers/issues/46


src/artifacts/mod.rs
```diff
@@ -791,19 +796,22 @@ pub enum EvmVersion {
     Berlin,
     London,
     Paris,
-    #[default]
     Shanghai,
+    #[default]
```
Vectorized, line 800: Should the default be Shanghai for now? To match the latest solc’s default.

mattsse (approved):
lgtm

Issue #34, opened 2023-12-18
Invalid contract flattening with matching interface names


SidestreamColdMelon:
This issue is the successor of https://github.com/gakonst/ethers-rs/issues/2588, created after the [deprecation announcement of ethers-rs](https://github.com/gakonst/ethers-rs/issues/2667).
Other issues relevant to flattening are collected here: https://github.com/foundry-rs/foundry/issues/6022.

### Problem

As previously described in https://github.com/foundry-rs/foundry/issues/4034, the latest version of forge (that now depends on `compilers` instead of `ethers-rs`) still doesn't properly handle situations when multiple files contain interfaces with the same name.

### Example

Given two files `FileA.sol` and `FileB.sol`

```solidity
// FileA.sol
pragma solidity ^0.8.10;
import "./FileB.sol";

interface FooBar {
    function foo() external;
}
contract A {
    function execute() external {
        FooBar(address(0)).foo();
    }
}
```

```solidity
// FileB.sol
pragma solidity ^0.8.10;

interface FooBar {
    function bar() external;
}
contract B {
    function execute() external {
        FooBar(address(0)).bar();
    }
}
```

Flattening will currently produce:

```solidity
pragma solidity ^0.8.10;

interface FooBar {
    function bar() external;
}
contract B {
    function execute() external {
        FooBar(address(0)).bar();
    }
}

interface FooBar { // here is the conflict with the interface declared above
    function foo() external;
}
contract A {
    function execute() external {
        FooBar(address(0)).foo();
    }
}
```

Instead, the desired outcome is:

```solidity
pragma solidity ^0.8.10;

interface FooBar_2 {
    function bar() external;
}
contract B {
    function execute() external {
        FooBar_2(address(0)).bar();
    }
}

interface FooBar_1 { // conflict avoided since the interface above was renamed before flattening
    function foo() external;
}
contract A {
    function execute() external {
        FooBar_1(address(0)).foo();
    }
}
```

### Notes

- Although it's obviously discussable, we specifically propose to use incremental `_N` suffix for all interfaces in order to keep flattening backward compatible with [`hevm flatten` command output](https://github.com/dapphub/dapptools/tree/master/src/hevm). It is also more readable and independent from the project structure compared to the potential path-based prefix

SidestreamColdMelon:
Closed by https://github.com/foundry-rs/compilers/pull/52

PR #52, opened 2024-01-26
New flattening impl

klkvr:
Ref https://github.com/foundry-rs/foundry/issues/4034.

## Solution
solang-parser AST was not sufficient to rename stuff, because it doesn't collect IDs of declarations and we can't match specific identifier to the declaration. Native solc AST gives such ability, so I've implemented native solc AST visitor.

The implementation of flattening itself consits of several steps:
1. Find all sources that are needed for target (imports of any depth).
2. Expore ASTs for all sources and find all top-level declarations (contracts, events, structs, etc).
3. Find if any top-level declarations names are same
4. If any duplicates are found, figure out new names for them (e.g. 2 `OracleLike` interfaces become `OracleLike_0` and `OracleLike_1`)
5. Walk AST and find all references to top-level declarations. Replace every reference with declaration name. We should update names even for unchanged declarations to deal with aliased imports.
6. Collect and remove all import directives.
7. Collect and remove all pragmas and license identifiers.

This flattener implementation can be a full replacer of the current impl for all cases when source being flattened can be compiled via solc.

### DIFF
--- a/src/flatten.rs
+++ b/src/flatten.rs
@@ -1,0 +1,398 @@
+use std::{
+    collections::{HashMap, HashSet},
+    hash::Hash,
+    path::{Path, PathBuf},
+};
+
+use crate::{
+    artifacts::{
+        ast::SourceLocation,
+        visitor::{Visitor, Walk},
+        Identifier, IdentifierPath, MemberAccess, Source, SourceUnit, SourceUnitPart, Sources,
+        UserDefinedTypeName,
+    },
+    error::SolcError,
+    utils, Graph, Project, ProjectCompileOutput, ProjectPathsConfig, Result,
+};
+
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
+struct ItemLocation {
+    path: PathBuf,
+    start: usize,
+    end: usize,
+}
+
+impl ItemLocation {
+    fn try_from_source_loc(src: &SourceLocation, path: PathBuf) -> Option<Self> {
+        if src.start.is_none() || src.length.is_none() {
+            return None;
+        }
+        let start = src.start.unwrap();
+        let end = start + src.length.unwrap();
+
+        Some(ItemLocation { path, start, end })
+    }
+}
+
+/// Visitor exploring AST and collecting all references to any declarations
+struct ReferencesCollector {
+    path: PathBuf,
+    references: HashMap<isize, HashSet<ItemLocation>>,
+}
+
+impl ReferencesCollector {
+    fn process_referenced_declaration(&mut self, id: isize, src: &SourceLocation) {
+        if let Some(loc) = ItemLocation::try_from_source_loc(src, self.path.clone()) {
+            self.references.entry(id).or_default().insert(loc);
+        }
+    }
+}
+
+impl Visitor for ReferencesCollector {
+    fn visit_identifier(&mut self, identifier: &Identifier) {
+        if let Some(id) = identifier.referenced_declaration {
+            self.process_referenced_declaration(id, &identifier.src);
+        }
+    }
+
+    fn visit_user_defined_type_name(&mut self, type_name: &UserDefinedTypeName) {
+        self.process_referenced_declaration(type_name.referenced_declaration, &type_name.src);
+    }
+
+    fn visit_member_access(&mut self, member_access: &MemberAccess) {
+        if let Some(id) = member_access.referenced_declaration {
+            self.process_referenced_declaration(id, &member_access.src);
+        }
+    }
+
+    fn visit_identifier_path(&mut self, path: &IdentifierPath) {
+        self.process_referenced_declaration(path.referenced_declaration, &path.src);
+    }
+}
+
+/// Context for flattening. Stores all sources and ASTs that are in scope of the flattening target.
+pub struct Flattener {
+    target: PathBuf,
+    sources: Sources,
+    asts: Vec<(PathBuf, SourceUnit)>,
+    // Sources in the order they should be written to the output file.
+    ordered_sources: Vec<PathBuf>,
+}
+
+impl Flattener {
+    pub fn new(project: &Project, output: &ProjectCompileOutput, target: &Path) -> Result<Self> {
+        // Performs DFS to collect all dependencies of a target
+        fn collect_deps(
+            path: &PathBuf,
+            paths: &ProjectPathsConfig,
+            graph: &Graph,
+            deps: &mut HashSet<PathBuf>,
+            ordered_deps: &mut Vec<PathBuf>,
+        ) -> Result<()> {
+            if deps.insert(path.clone()) {
+                let target_dir = path.parent().ok_or_else(|| {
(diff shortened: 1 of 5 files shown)


klkvr:
This should also fix issues with aliases (like https://github.com/foundry-rs/foundry/issues/2545, https://github.com/foundry-rs/foundry/issues/3265). 

Current impl handles aliases renaming via regex'es and may often be inaccurate.

src/flatten.rs
```diff
@@ -0,0 +1,398 @@
...
+                let target_dir = path.parent().ok_or_else(|| {
+                    SolcError::msg(format!(
+                        "failed to get parent directory for \"{:?}\"",
+                        path.display()
+                    ))
+                })?;
+
+                let node_id = graph.files().get(path).ok_or_else(|| {
+                    SolcError::msg(format!("cannot resolve file at {:?}", path.display()))
+                })?;
```
DaniPopes, line 102: use either `{}, path.display()` or `{:?}, path`

klkvr:
I've refactored code and added logic for correct processing of references to types declared in scope of contracts.

It requires such workaround because for some reason solc does not give any information about the parent contract for `UserDefinedTypeName` objects representing types like `ParentContract.EnumName`. Such node will only have AST ID and location of `EnumName` definition included.

That way, if we want to rename contract `ParentContract`, we need to be able to determine that AST ID of `EnumName` is related to that contract and replace it with `ParentContract_i.EnumName`

src/flatten.rs
```diff
@@ -0,0 +1,561 @@
...
+    /// Flattens target file and returns the result as a string
+    ///
+    /// Flattening process includes following steps:
+    /// 1. Find all file-level definitions and rename references to them via aliased or qualified
+    ///    imports.
+    /// 2. Find all duplicates among file-level definitions and rename them to avoid conflicts.
+    /// 3. Remove all imports.
+    /// 4. Remove all pragmas except for the ones in the target file.
+    /// 5. Remove all license identifiers except for the one in the target file.
+    pub fn flatten(&self) -> String {
```
mattsse, line 179: can we split this into a two step process,

one additional (internal) struct that wraps Updates + Flattener, basically `let updated_sources = self.get_updated_sources(updates)`

that then has function to construct the output string. this could perhaps come in handy for some tests

mattsse (changes requested):
this looks great!!!

can add some additional tests for #34 ?

mattsse:
@sakulstra I believe you had some issues with flattening in the past, do you know any projects we should test this against?

sakulstra:
@mattsse there were a few issues/things i commented in the past:
- https://github.com/foundry-rs/foundry/issues/2378
e.g. `cast etherscan-source -d etherscan/0x7b2a3cf972c3193f26cdec6217d27379b6417bd0 0x7b2a3cf972c3193f26cdec6217d27379b6417bd0 && forge flatten ./etherscan/0x7b2a3cf972c3193f26cdec6217d27379b6417bd0/AToken/@aave/protocol-v2/contracts/protocol/tokenization/AToken.sol --output ./etherscan/0x7b2a3cf972c3193f26cdec6217d27379b6417bd0/Flattened.sol` would remove `pragma experimental ABIEncoderV2;` and therefore creating non-identical(in this case breaking) code.

- https://github.com/foundry-rs/foundry/issues/6539 not really an issue but more a feature request i guess?

For #4034 specifically we also had some issues with that, but can't recall right now in which contracts. Will check if i can find on monday.

klkvr:
Added test and fix for https://github.com/foundry-rs/foundry/issues/2378

klkvr:
Pushed solution for https://github.com/foundry-rs/foundry/issues/6539.

https://github.com/foundry-rs/compilers/blob/ce8e42746599694bb884f06e69d5014e5d7a10e8/src/flatten.rs#L586-L624

 I've also shared it's logic with current flattening implementation and rewritten it a bit as it's not longer need to be recursive.

src/flatten.rs
```diff
@@ -0,0 +1,623 @@
...
+    end: usize,
+}
+
+impl ItemLocation {
+    fn try_from_source_loc(src: &SourceLocation, path: PathBuf) -> Option<Self> {
+        if src.start.is_none() || src.length.is_none() {
+            return None;
+        }
+        let start = src.start.unwrap();
+        let end = start + src.length.unwrap();
```
mattsse, line 31: use `?` operator here

src/flatten.rs
```diff
@@ -0,0 +1,623 @@
...
+    pub fn flatten(&self) -> String {
+        let mut updates = Updates::new();
+
+        let top_level_names = self.rename_top_level_definitions(&mut updates);
+        self.rename_contract_level_types_references(&top_level_names, &mut updates);
+        self.remove_imports(&mut updates);
+        let target_pragmas = self.process_pragmas(&mut updates);
+        let target_license = self.process_licenses(&mut updates);
+
+        let result = FlatteningResult::new(self, updates, target_pragmas, target_license);
```
mattsse, line 215: FlattenedResult only borrows from `Flattener`, right?

let's move this to `fn flatten_result` so that this function becomes `self.flatten_result().get_flattened_target`
if possible

mattsse (approved):
nice, last cosmetic nits!

klkvr:
@mattsse pushed fixes

let's merge a bit later, I will now write integration of it into foundry and may decide to change the API of `Flattener` a little bit

mattsse (approved):
sgtm!
pending foundry companion pr

src/config.rs
```diff
@@ -401,110 +402,95 @@ impl ProjectPathsConfig {
...
+                    if let Some(name_match) =
+                        ["n1", "n2", "n3"].iter().find_map(|name| cap.name(name))
+                    {
+                        let name_match_range =
+                            utils::range_by_offset(&name_match.range(), replace_offset);
+                        replace_offset += target_len - (name_match_range.len() as isize);
+                        content.replace_range(name_match_range, &target);
+                    }
+                }
+            }
```
Evalir, line 444: could we add a bit of docs over what's happening here?

src/flatten.rs
```diff
@@ -0,0 +1,627 @@
...
+        ast::SourceLocation,
+        visitor::{Visitor, Walk},
+        ContractDefinitionPart, Identifier, IdentifierPath, MemberAccess, Source, SourceUnit,
+        SourceUnitPart, Sources, UserDefinedTypeName,
+    },
+    error::SolcError,
+    utils, Graph, Project, ProjectCompileOutput, ProjectPathsConfig, Result,
+};
+
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
```
Evalir, line 18: can we add some docs here?

Evalir (approved):
amazing. some nits on docs only

src/config.rs
klkvr: will add, however this code was written before me so I might get something wrong haha

PR #57, opened 2024-01-30
chore: add more tracing around spawning Solc

DaniPopes:


src/compile/mod.rs
```diff
@@ -546,66 +546,88 @@ impl Solc {
...
+        let stdin = child.stdin.as_mut().unwrap();
         serde_json::to_writer(stdin, input)?;
-        compile_output(child.wait_with_output().map_err(|err| SolcError::io(err, &self.solc))?)
+        debug!("wrote JSON input to stdin");
+
+        let output = child.wait_with_output().map_err(self.map_io_err())?;
+        trace!(
+            output.stdout = ?String::from_utf8_lossy(&output.stdout),
+            output.stderr = ?String::from_utf8_lossy(&output.stderr),
+        );
```
mattsse, line 603: this can be gigantic
often > 100MB

bug I guess we could really use this

mattsse (approved):

PR #58, opened 2024-01-30
refactor: rewrite examples without wrapper functions and with no_run

DaniPopes:
Using a wrapper function like `fn f(x: _) -> Result` hides the fact that all the code in the example does not run

Also use `?` instead of unwrap


mattsse (approved):
ty boss 🫡

PR #63, opened 2024-01-30
Flatten fixes

klkvr:
Support for more flattening edge cases.

Resolves https://github.com/foundry-rs/foundry/pull/6936#issuecomment-1916511765 and https://github.com/foundry-rs/foundry/pull/6936#issuecomment-1915222867

We still don't rename stuff in yul blocks.

prb-math turned out to be a great codebase to test flattening against as it has a lot of unusual patterns (qulified imports, user-defined functions on user-defined types, etc). Currently, their contracts flattened by this flattener are only having troubles with yul code blocks

### DIFF
--- a/src/flatten.rs
+++ b/src/flatten.rs
@@ -9,7 +9,7 @@ use crate::{
         ast::SourceLocation,
         visitor::{Visitor, Walk},
         ContractDefinitionPart, Identifier, IdentifierPath, MemberAccess, Source, SourceUnit,
-        SourceUnitPart, Sources, UserDefinedTypeName,
+        SourceUnitPart, Sources,
     },
     error::SolcError,
     utils, Graph, Project, ProjectCompileOutput, ProjectPathsConfig, Result,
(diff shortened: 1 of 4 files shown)


src/flatten.rs
```diff
@@ -53,32 +65,22 @@ impl Visitor for ReferencesCollector {
...
-impl Visitor for UserDefinedTypeNamesCollector {
-    fn visit_user_defined_type_name(&mut self, type_name: &UserDefinedTypeName) {
-        if let Some(loc) = ItemLocation::try_from_source_loc(&type_name.src, self.path.clone()) {
-            self.references.entry(type_name.referenced_declaration).or_default().insert(loc);
+    fn visit_member_access(&mut self, access: &MemberAccess) {
+        if let Some(referenced_declaration) = access.referenced_declaration {
+            let length = access.member_name.len();
+            // Accessed member name is in the last name.len() symbols of the expression.
+            let start = access.src.start.unwrap() + access.src.length.unwrap() - length;
+            let end = start + length;
```
mattsse, line 77: this should be safe, right?

mattsse (approved):

src/flatten.rs
klkvr: It's safe for latest versions, however thinking of it now it might be better to make it a bit safer as I'm not sure what might be happening on older versions

will push a fix

mattsse (approved):

DaniPopes (approved):

PR #68, opened 2024-02-01
Flatten fix

klkvr:
Ref https://github.com/foundry-rs/foundry/pull/6936#issuecomment-1921443885

Not sure why exactly this unwrap is failing, adding `None` variant handling and some logs to debug it.

### DIFF
--- a/src/flatten.rs
+++ b/src/flatten.rs
@@ -436,8 +436,8 @@ impl Flattener {
                 let content: &str = &self.sources.get(path).unwrap().content[src_start..src_end];
                 let tag_len = "@inheritdoc".len();
 
-                trace!("processing doc with content {:?}", content);
                 if let Some(tag_start) = content.find("@inheritdoc") {
+                    trace!("processing doc with content {:?}", content);
                     if let Some(name_start) = content[tag_start + tag_len..]
                         .find(|c| c != ' ')
                         .map(|p| p + tag_start + tag_len)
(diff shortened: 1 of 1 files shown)


src/flatten.rs
```diff
@@ -445,20 +448,36 @@ impl Flattener {
...
+                            if let Some(name) = top_level_names.get(ast_id) {
+                                new_name = Some(name);
+                            } else {
+                                trace!("ast id {ast_id} cannot be matched to top-level identifier");
+                                trace!("known top-level identifiers: {:?}", top_level_names);
+                            }
+                        }
+
+                        match new_name {
+                            Some(new_name) => {
```
mattsse, line 465: use if let Some else

src/flatten.rs
```diff
@@ -445,20 +448,36 @@ impl Flattener {
...
-                            updates.entry(path.to_path_buf()).or_default().insert((
-                                src_start + tag_start,
-                                src_start + name_end,
-                                "".to_string(),
-                            ));
+                            if let Some(name) = top_level_names.get(ast_id) {
+                                new_name = Some(name);
+                            } else {
+                                trace!("ast id {ast_id} cannot be matched to top-level identifier");
+                                trace!("known top-level identifiers: {:?}", top_level_names);
```
mattsse, line 460: these can be combined
```suggestion
                                trace!(identifiers=?top_level_names, "ast id {ast_id} cannot be matched to top-level identifier");
```

mattsse (changes requested):
pedantic nits

mattsse (approved):

PR #69, opened 2024-02-05
Option to ignore warnings from dependencies in foundry.toml

loocapro:
[Related to #1200](https://github.com/foundry-rs/foundry/issues/1200)

### DIFF
--- a/src/artifacts/mod.rs
+++ b/src/artifacts/mod.rs
@@ -1572,7 +1572,8 @@ impl CompilerOutput {
 
     /// Checks if there are any compiler warnings that are not ignored by the specified error codes
     /// and file paths.
-    pub fn has_warning(&self, filter: ErrorFilter<'_>) -> bool {
+    pub fn has_warning<'a>(&self, filter: impl Into<ErrorFilter<'a>>) -> bool {
+        let filter: ErrorFilter<'_> = filter.into();
         self.errors.iter().any(|error| {
             if !error.severity.is_warning() {
                 return false;
--- a/src/artifacts/mod.rs
+++ b/src/artifacts/mod.rs
@@ -1571,10 +1571,22 @@ impl CompilerOutput {
     }
 
     /// Whether the output contains a compiler warning
-    pub fn has_warning(&self, ignored_error_codes: &[u64]) -> bool {
+    pub fn has_warning(&self, ignored_error_codes: &[u64], ignored_file_paths: &[PathBuf]) -> bool {
         self.errors.iter().any(|err| {
             if err.severity.is_warning() {
-                err.error_code.as_ref().map_or(false, |code| !ignored_error_codes.contains(code))
+                let is_code_not_ignored = err
+                    .error_code
+                    .as_ref()
+                    .map_or(false, |code| !ignored_error_codes.contains(code));
+
+                let is_file_not_ignored = err.source_location.as_ref().map_or(false, |location| {
+                    !ignored_file_paths.contains(&PathBuf::from(&location.file))
+                });
+
+                // Return true if the error code is not ignored and the file path is not ignored
+                // This means the error is considered a warning if it's not excluded by either the
+                // ignored codes or paths
+                is_code_not_ignored && is_file_not_ignored
             } else {
                 false
             }
--- a/src/artifacts/mod.rs
+++ b/src/artifacts/mod.rs
@@ -1577,22 +1577,22 @@ impl CompilerOutput {
             if !error.severity.is_warning() {
                 return false;
             }
-            let is_code_not_ignored =
-                error.error_code.map_or(false, |code| !ignored_error_codes.contains(&code));
+            let is_code_ignored =
+                error.error_code.map_or(false, |code| ignored_error_codes.contains(&code));
 
-            let is_file_not_ignored = self.is_file_not_ignored(error, ignored_file_paths);
-            is_code_not_ignored && is_file_not_ignored
+            let is_file_ignored = self.is_file_ignored(error, ignored_file_paths);
+            is_code_ignored || is_file_ignored
         })
     }
 
     /// Determines if the error's file path is not ignored based on a list of ignored file paths.
-    fn is_file_not_ignored(&self, error: &Error, ignored_file_paths: &[PathBuf]) -> bool {
+    fn is_file_ignored(&self, error: &Error, ignored_file_paths: &[PathBuf]) -> bool {
         match &error.source_location {
             Some(location) => {
                 // Convert the error's file location to a PathBuf for comparison
                 let error_path = PathBuf::from(&location.file);
                 // Check if the error path is not in the list of ignored paths
-                !ignored_file_paths.iter().any(|ignored_path| error_path.starts_with(ignored_path))
+                ignored_file_paths.iter().any(|ignored_path| error_path.starts_with(ignored_path))
             }
             None => false,
         }
--- a/src/artifacts/mod.rs
+++ b/src/artifacts/mod.rs
@@ -1580,7 +1580,16 @@ impl CompilerOutput {
                     .map_or(false, |code| !ignored_error_codes.contains(code));
 
                 let is_file_not_ignored = err.source_location.as_ref().map_or(false, |location| {
-                    !ignored_file_paths.contains(&PathBuf::from(&location.file))
+                    // Convert ignored_file_paths to a Vec<String> for easier comparison
+                    let ignored_paths_as_strings: Vec<String> = ignored_file_paths
+                        .iter()
+                        .map(|path| path.to_string_lossy().into_owned())
+                        .collect();
+
+                    // Check if location.file contains any of the ignored_file_paths elements
+                    !ignored_paths_as_strings
+                        .iter()
+                        .any(|ignored_path| location.file.contains(ignored_path))
                 });
 
                 // Return true if the error code is not ignored and the file path is not ignored
--- a/src/artifacts/mod.rs
+++ b/src/artifacts/mod.rs
@@ -1570,7 +1570,6 @@ impl CompilerOutput {
         self.errors.iter().any(|err| err.severity.is_error())
     }
 
-    /// Whether the output contains a compiler warning
     /// Checks if there are any compiler warnings that are not ignored by the specified error codes
     /// and file paths.
     pub fn has_warning(&self, ignored_error_codes: &[u64], ignored_file_paths: &[PathBuf]) -> bool {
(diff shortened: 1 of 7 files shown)


src/compile/output/mod.rs
```diff
@@ -32,6 +33,8 @@ pub struct ProjectCompileOutput<T: ArtifactOutput = ConfigurableArtifacts> {
     pub(crate) cached_artifacts: Artifacts<T::Artifact>,
     /// errors that should be omitted
     pub(crate) ignored_error_codes: Vec<u64>,
+    /// paths that should be omitted
+    pub(crate) ignored_file_paths: ProjectPaths,
```
mattsse, line 37: I think we either want a Vec<PathBuf> or an Option here

because now it's not obvious to distinguish between "include library errors" or exclude them

the Vec<PathBuf> solution is likely more flexible 

if configured, all warnings that originate in a path will be skipped

mattsse (changes requested):

src/compile/output/mod.rs
```diff
@@ -529,7 +550,7 @@ impl AggregatedCompilerOutput {
     pub fn diagnostics<'a>(
         &'a self,
         ignored_error_codes: &'a [u64],
-        ignored_file_paths: &'a ProjectPaths,
+        ignored_file_paths: &'a Vec<PathBuf>,
```
mattsse, line 558:
```suggestion
        ignored_file_paths: &'a [PathBuf],
```

src/compile/output/mod.rs
```diff
@@ -798,20 +819,24 @@ pub struct OutputDiagnostics<'a> {
     /// the error codes to ignore
     ignored_error_codes: &'a [u64],
     /// the file paths to ignore
-    ignored_file_paths: &'a ProjectPaths,
+    ignored_file_paths: &'a Vec<PathBuf>,
```
mattsse, line 827:
```suggestion
    ignored_file_paths: &'a [PathBuf],
```

src/compile/output/mod.rs
```diff
@@ -516,10 +525,22 @@ impl AggregatedCompilerOutput {
...
         self.errors.iter().any(|err| {
             if err.severity.is_warning() {
-                err.error_code.as_ref().map_or(false, |code| !ignored_error_codes.contains(code))
+                let is_code_not_ignored = err
+                    .error_code
+                    .as_ref()
+                    .map_or(false, |code| !ignored_error_codes.contains(code));
+
+                let is_file_not_ignored = err.source_location.as_ref().map_or(false, |location| {
+                    !ignored_file_paths.contains(&PathBuf::from(&location.file))
```
mattsse: this should not be `contains`, instead we want to check if the file is inside one fo the ignored files (`starts_with`)

src/compile/output/mod.rs
```diff
@@ -516,10 +525,22 @@ impl AggregatedCompilerOutput {
...
+    pub fn has_warning(&self, ignored_error_codes: &[u64], ignored_file_paths: &[PathBuf]) -> bool {
         self.errors.iter().any(|err| {
             if err.severity.is_warning() {
-                err.error_code.as_ref().map_or(false, |code| !ignored_error_codes.contains(code))
+                let is_code_not_ignored = err
+                    .error_code
+                    .as_ref()
+                    .map_or(false, |code| !ignored_error_codes.contains(code));
+
+                let is_file_not_ignored = err.source_location.as_ref().map_or(false, |location| {
```
mattsse: this needs to check OR:

if is_error_ignored || is_file_ignored -> return false

src/compile/output/mod.rs
```diff
@@ -516,10 +525,22 @@ impl AggregatedCompilerOutput {
     }
 
     /// Whether the output contains a compiler warning
-    pub fn has_warning(&self, ignored_error_codes: &[u64]) -> bool {
+    pub fn has_warning(&self, ignored_error_codes: &[u64], ignored_file_paths: &[PathBuf]) -> bool {
```
mattsse, line 529: I think we want a helper type for this, in case we want to fine-tune this in the future, for example "ignore these errors in these files"

I'd like something like

```rust
struct ErrorFilter<'a> {
  error_codes: Cow<'a, [u64]>,
  ignored_file_paths: Cow<'a, [PathBuf]>,
}
```

and a bunch of helper functions and From impls

so we can try to:

```rust
fn has_warning<'a>(&self, filter: impl Into<ErrorFilter<'a>)
```
perhaps, maybe not possible due to lifetime

mattsse (changes requested):
nice, this is pretty close.

only have some style nitpicks

src/artifacts/mod.rs
```diff
@@ -1570,17 +1570,34 @@ impl CompilerOutput {
...
+            is_code_ignored || is_file_ignored
         })
     }
 
+    /// Determines if the error's file path is not ignored based on a list of ignored file paths.
+    fn is_file_ignored(&self, error: &Error, ignored_file_paths: &[PathBuf]) -> bool {
+        match &error.source_location {
+            Some(location) => {
+                // Convert the error's file location to a PathBuf for comparison
+                let error_path = PathBuf::from(&location.file);
```
DaniPopes, line 1593: Path::new

src/compile/output/mod.rs
```diff
@@ -498,6 +502,36 @@ pub struct AggregatedCompilerOutput {
...
+        }
+    }
+
+    // Helper function to check if an error's file path is ignored
+    pub(crate) fn is_file_ignored(&self, file_path: &Path) -> bool {
+        self.ignored_file_paths.iter().any(|ignored_path| file_path.starts_with(ignored_path))
+    }
+}
+
+impl<'a> From<(&'a [u64], &'a [PathBuf])> for ErrorFilter<'a> {
```
mattsse, line 523: this also needs From<&[u64]> for backwards compat

src/compile/output/mod.rs
```diff
@@ -526,30 +561,24 @@ impl AggregatedCompilerOutput {
...
-        match &error.source_location {
-            Some(location) => {
-                // Convert the error's file location to a PathBuf for comparison
-                let error_path = PathBuf::from(&location.file);
-                // Check if the error path is not in the list of ignored paths
-                !ignored_file_paths.iter().any(|ignored_path| error_path.starts_with(ignored_path))
-            }
-            None => false,
-        }
+            let is_file_ignored = match &error.source_location {
```
mattsse, line 570: no match if Option, use map_or or .map().unwrap_or_default

mattsse (changes requested):
almost there :)

mattsse (changes requested):
nice, let's add a simple test for this

todo:

add another test similar to this:

https://github.com/foundry-rs/compilers/blob/564eb8591adde1e5a33e9d07a7d18b8f59a8fcdb/tests/project.rs#L1846-L1846

The lib code should trigger a warning

configure ignored paths and check output diagnostic

src/artifacts/mod.rs
```diff
@@ -16,7 +17,6 @@ use std::{
     str::FromStr,
     sync::Arc,
 };
-
```
DaniPopes, line 19: pls revert

src/compile/output/mod.rs
```diff
@@ -484,6 +500,45 @@ pub struct AggregatedCompilerOutput {
...
+        }
+    }
+
+    // Helper function to check if an error's file path is ignored
+    pub(crate) fn is_file_ignored(&self, file_path: &Path) -> bool {
+        self.ignored_file_paths.iter().any(|ignored_path| file_path.starts_with(ignored_path))
+    }
+}
+
+impl<'a> From<(&'a [u64], &'a [PathBuf])> for ErrorFilter<'a> {
```
DaniPopes, line 523: I'd prefer if this was a `fn new` constructor with `impl Into<Cow>`. I find `From` constructors not idiomatic or intuitive

src/compile/output/mod.rs
```diff
@@ -484,6 +500,45 @@ pub struct AggregatedCompilerOutput {
     pub build_infos: BTreeMap<Version, RawBuildInfo>,
 }
 
+pub struct ErrorFilter<'a> {
+    error_codes: Cow<'a, [u64]>,
+    ignored_file_paths: Cow<'a, [PathBuf]>,
```
DaniPopes, line 505: fields should be public

src/artifacts/mod.rs
```diff
@@ -1570,14 +1570,26 @@ impl CompilerOutput {
...
+            if !error.severity.is_warning() {
+                return false;
             }
+
+            let is_code_ignored = filter.is_code_ignored(error.error_code);
+
+            let is_file_ignored = error
+                .source_location
+                .as_ref()
+                .map_or(false, |location| filter.is_file_ignored(&PathBuf::from(&location.file)));
```
DaniPopes, line 1587: Path::new

src/compile/output/mod.rs
```diff
@@ -833,6 +914,11 @@ impl<'a> fmt::Display for OutputDiagnostics<'a> {
                         // from a test file we skip
                         ignored =
                             self.is_test(&source_location.file) && (code == 1878 || code == 5574);
+
+                        // we ignore warnings coming from ignored files
+                        ignored |= self
+                            .ignored_file_paths
+                            .starts_with(&[PathBuf::from(&source_location.file)]);
```
DaniPopes, line 921: Path::new, also I don't think this is right? Why `starts_with` on a list of ignored paths?

tests/project.rs
```diff
@@ -17,6 +17,7 @@ use foundry_compilers::{
     ProjectCompileOutput, ProjectPathsConfig, Solc, TestFileFilter,
 };
 use pretty_assertions::assert_eq;
+use rayon::vec;
```
DaniPopes, line 20: I assume this is a mistake

DaniPopes (changes requested):

src/compile/output/mod.rs
```diff
@@ -484,6 +500,42 @@ pub struct AggregatedCompilerOutput {
     pub build_infos: BTreeMap<Version, RawBuildInfo>,
 }
 
+pub struct ErrorFilter<'a> {
+    pub error_codes: Cow<'a, [u64]>,
+    pub ignored_file_paths: Cow<'a, [PathBuf]>,
+}
+
+impl<'a> ErrorFilter<'a> {
```
DaniPopes, line 508: All of these should be public

DaniPopes (approved):

src/compile/output/mod.rs
mattsse: same,

main motivation for this was so we don't need to break if providing &[u64] error codes still works.

but the tuple we should actually remove yes

src/compile/output/mod.rs
mattsse: it should be the other way around yes

mattsse (approved):
lgtm

PR #70, opened 2024-02-07
fix: cleanup build info path

0xalpharush:
## Motivation
The build info path was not being cleaned and leaving out-of-date build info on disk causing issues reconciling the source code with solc AST 

xref https://github.com/crytic/slither/issues/2296

## Solution
Delete build info path


mattsse:
sorry, I only saw the ethers pr

superseded by #71

PR #71, opened 2024-02-07
fix: also cleanup build info dir

mattsse:
ref https://github.com/gakonst/ethers-rs/pull/2730


DaniPopes:
#70 ?

mattsse:
agh sigh, overlooked this one -.-

let's more this one because I added a simple test

PR #72, opened 2024-02-07
Helper method for `Libraries`

klkvr:
Adds helper method to strip path prefixes from `Libraries` object, I'll need this for https://github.com/foundry-rs/foundry/pull/7027


onbjerg (approved):
lgtm

mattsse (approved):

PR #75, opened 2024-02-13
Small flattener features

klkvr:
This PR implements two small flattener features mentioned in https://github.com/makerdao/spells-mainnet/pull/391#issuecomment-1926898173

1. Combine pragmas from different sources into one pragma so that version restrictions are kept the same before and after flattening
2. Comments specifying original source file of the flattened code (`// src/File.sol`)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ home = "0.5"
 svm = { package = "svm-rs", version = "0.3", default-features = false, optional = true }
 svm-builds = { package = "svm-rs-builds", version = "0.3", default-features = false, optional = true }
 sha2 = { version = "0.10", default-features = false, optional = true }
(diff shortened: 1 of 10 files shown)


Cargo.toml
```diff
@@ -57,6 +57,7 @@ home = "0.5"
 svm = { package = "svm-rs", version = "0.3", default-features = false, optional = true }
 svm-builds = { package = "svm-rs-builds", version = "0.3", default-features = false, optional = true }
 sha2 = { version = "0.10", default-features = false, optional = true }
+itertools = "0.11"
```
mattsse, line 60:
```suggestion
itertools = "0.12"
```

mattsse (approved):
lgtm

onbjerg (approved):

PR #85, opened 2024-02-22
fix: account for Solc inexplicably not formatting the message

DaniPopes:
Apparently it's not just old solc versions, some errors just are not formatted on the latest versions too.

Edit: if the file name does not end in .sol, the source locations are not formatted. Lol


mattsse (approved):

DaniPopes:
Testing with https://github.com/foundry-rs/foundry/pull/7210

PR #87, opened 2024-03-07
feat: use cached artifacts if solc config is almost the same

klkvr:
If the only difference between cached vs current solc config is `outputSelection` which is a subset of the cached output selection, then it's safe to use cached artifact.

I am currently adding logic to foundry to request only abi output and it would be nice if we could use cached abi where possible.

Impl is pretty ugly, but not sure how to make it better without cloning

### DIFF
--- a/src/artifacts/output_selection.rs
+++ b/src/artifacts/output_selection.rs
@@ -119,6 +119,19 @@ impl OutputSelection {
     pub fn empty_file_output_select() -> FileOutputSelection {
         Default::default()
     }
+
+    /// Returns true if this output selection is a subset of the other output selection.
+    pub fn is_subset_of(&self, other: &Self) -> bool {
+        self.0.iter().all(|(file, selection)| {
+            other.0.get(file).map_or(false, |other_selection| {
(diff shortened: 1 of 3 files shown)


src/artifacts/output_selection.rs
```diff
@@ -119,6 +119,19 @@ impl OutputSelection {
...
+    pub fn is_subset_of(&self, other: &Self) -> bool {
+        self.0.iter().all(|(file, selection)| {
+            other.0.get(file).map_or(false, |other_selection| {
+                selection.iter().all(|(contract, outputs)| {
+                    other_selection.get(contract).map_or(false, |other_outputs| {
+                        outputs.iter().all(|output| other_outputs.contains(output))
+                    })
+                })
+            })
+        })
```
mattsse, line 133: can we add a sanity test for this?

looks correct though

mattsse (changes requested):
this makes a lot of sense,

smol sanity test then this is gtg

mattsse (approved):
lgtm!!

PR #88, opened 2024-03-08
refactor: extra files logic

klkvr:
Resolves https://github.com/foundry-rs/foundry/issues/6241 

Removes `write_extra` logic from `ArtifactOutput` trait
Adds new methods:
- `handle_artifacts` - by default empty implementation which can be overloaded to implement handling of compiled artifacts. `ConfigurableArtifacts` implements it to write extra files
- `is_dirty` - method allowing `ArtifactOutput` implementations to reject cached artifacts and enforce recompilation. Used by `ConfigurableArtifacts` to reject artifacts which data is not enough to write required extra output files.
- `handle_cached_artifacts` - same as `handle_artifacts` but for those which were not recompiled. Used by `ConfigurableArtifacts` to write extra files from cached artifacts.

### DIFF
--- a/src/artifact_output/configurable.rs
+++ b/src/artifact_output/configurable.rs
@@ -784,7 +784,6 @@ impl ExtraOutputFiles {
     /// Write the set values as separate files
     pub fn write_extras(&self, contract: &Contract, file: &Path) -> Result<(), SolcError> {
         self.process_abi(contract.abi.as_ref(), file)?;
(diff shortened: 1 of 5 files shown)


mattsse (approved):
this is a far better solution!

holic:
thank you!

curious if this also resolves the issue in this comment: https://github.com/foundry-rs/foundry/issues/6241#issuecomment-1802142594

happy to open a new issue if not

PR #90, opened 2024-03-12
refactor: caching logic

klkvr:
Currently we are not keeping cache entries for files which were out of scope of compiler (might be dirty or not)

Because of that, when running `forge compile`, then running `forge compile --skip ...` which will filter some artifacts out, and then running `forge compile` again, some artifacts will be recompiled, because `--skip` run removed those entries from cache.

Solution is to keep those entries in cache (we already silently keep artifacts anyway).

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -476,7 +476,7 @@ impl CacheEntry {
             for artifact in artifacts {
                 self.artifacts
                     .entry(name.clone())
-                    .or_insert_with(BTreeMap::new)
+                    .or_default()
                     .insert(artifact.version.clone(), artifact.file.clone());
             }
         }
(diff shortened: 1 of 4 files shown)


src/compile/project.rs
```diff
@@ -708,9 +704,8 @@ mod tests {
         let prep = compiler.preprocess().unwrap();
         let cache = prep.cache.as_cached().unwrap();
         // 3 contracts
-        assert_eq!(cache.dirty_source_files.len(), 3);
-        assert!(cache.filtered.is_empty());
-        assert!(cache.cache.is_empty());
```
klkvr, line 713: we are now adding entries for dirty files to cache during preprocessing

onbjerg (approved):
makes sense to me, ptal @mattsse

src/cache.rs
```diff
@@ -580,6 +553,28 @@ impl CacheEntry {
     }
 }
 
+#[derive(Debug, Clone, Default)]
+pub struct GroupedSources {
+    pub inner: HashMap<PathBuf, HashSet<Version>>,
+}
+
+impl GroupedSources {
+    pub fn insert(&mut self, file: PathBuf, version: Version) {
```
onbjerg, line 562: nit: docs for these

mattsse (approved):
great!
this makes a lot of sense

onbjerg (approved):

klkvr:
Patched #7334 with this PR branch, will merge after it succeeds

mattsse (approved):
this looks great!
clean

onbjerg (approved):
lgtm as well :shipit:

PR #100, opened 2024-04-13
feat: optimization field (simpleCounterForLoopUncheckedIncrement)

### LINKED ISSUE #99: Missing Optimization Field (simpleCounterForLoopUncheckedIncrement) in Foundry Configuration
In the Solidity [JSON input](https://docs.soliditylang.org/en/latest/using-the-compiler.html#input-description:~:text=//%20Use%20unchecked%20arithmetic%20when%20incrementing%20the%20counter%20of%20for%20loops%0A%20%20%20%20%20%20%20%20//%20under%20certain%20circumstances.%20It%20is%20always%20on%20if%20no%20details%20are%20given.%0A%20%20%20%20%20%20%20%20%22simpleCounterForLoopUncheckedIncrement%22%3A%20true%2C), a compilation optimization flag `simpleCounterForLoopUncheckedIncrement` is added to indicate using unchecked arithmetic when incrementing the counter of for loops under certain circumstances. However, in the current [codebase](https://github.com/foundry-rs/compilers/blob/899f9a0357d68de73e36b57deb01ebb84713141e/src/artifacts/mod.rs#L717), this field is missing.

This issue is also related to the PR mentioned in https://github.com/foundry-rs/foundry/pull/7576#discussion_r1562365875.

kamuik16:
Closes #99

### DIFF
--- a/src/artifacts/mod.rs
+++ b/src/artifacts/mod.rs
@@ -491,6 +491,7 @@ impl Settings {
                 // with only unused prunner step
                 optimizer_steps: Some("u".to_string()),
             }),
+            simple_counter_for_loop_unchecked_increment: Some(true),
         });
         self
     }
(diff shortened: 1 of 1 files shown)


src/artifacts/mod.rs
```diff
@@ -491,6 +491,7 @@ impl Settings {
                 // with only unused prunner step
                 optimizer_steps: Some("u".to_string()),
             }),
+            simple_counter_for_loop_unchecked_increment: Some(true),
```
mattsse, line 494: this makes sense, default is true, and we can set it

mattsse (approved):
lgtm, ty

PR #103, opened 2024-04-15
feat: additional helpers for contract name -> path lookup

klkvr:
ref https://github.com/foundry-rs/foundry/pull/7599


mattsse (changes requested):
can we make these functions of Project directly? so we have `fn(&self)`

mattsse (approved):

PR #104, opened 2024-04-16
fix: invalidate cache for out-of-scope entries

klkvr:
missed this in #90

ref https://t.me/foundry_support/52331

after #90 we stopped removing out of scope sources from cache, however, we never validated if they are dirty, this results in invalid cached artifacts in certain cases.

with this PR we iterate over all cache entries out of scope of current compiler run, then build graph with them to find if they import any of files marked as dirty.

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -6,7 +6,7 @@ use crate::{
     error::{Result, SolcError},
     filter::{FilteredSources, SourceCompilationKind},
     resolver::GraphEdges,
-    utils, ArtifactFile, ArtifactOutput, Artifacts, ArtifactsMap, OutputContext, Project,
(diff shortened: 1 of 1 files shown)


src/cache.rs
```diff
@@ -705,7 +705,38 @@ impl<'a, T: ArtifactOutput> ArtifactsCacheInner<'a, T> {
...
+                dirty_files.insert(cache_file.clone());
+                continue;
+            };
+            sources_out_of_scope.insert(cache_file.clone(), source);
+        }
+
+        let files = sources_out_of_scope.keys().cloned().collect::<Vec<_>>();
+
+        // Build a temporary graph to repeat the process above. We need this because `self.edges`
+        // only contains graph data for in-scope sources.
```
mattsse, line 730: ah, yeah, that makes sense

mattsse (approved):
nice,
triggering patch once merged

PR #105, opened 2024-04-17
fix: dirty files detection

klkvr:
Follow-up to #104. Right now we detect and remove dirty cache entries in `ArtifactsCacheInner::filter` which is redundant as we can just do a single iteration over all cached files.

Now file filtering is done in 2 steps:
1. `ArtifactsCacheInner::find_and_remove_dirty` detects all dirty entries (even those that are not in scope of current compiler run) and removes them from cache.
2. `ArtifactsCacheInner::filter` schedules for compilation all sources for which we are missing artifacts (either because of removed entry or because we never compiled given source with current version)

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -683,6 +683,7 @@ impl<'a, T: ArtifactOutput> ArtifactsCacheInner<'a, T> {
         false
     }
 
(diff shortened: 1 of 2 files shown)


mattsse (approved):
lgtm

PR #107, opened 2024-04-19
Update yansi to 1.0

dcb9:
Ref https://github.com/foundry-rs/foundry/issues/7709

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ serde_json = "1.0"
 thiserror = "1.0"
 tracing = "0.1"
 walkdir = "2.4"
-yansi = "0.5"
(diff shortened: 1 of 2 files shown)


mattsse (changes requested):
looking at some docs

> // The following two are equivalent. The latter is preferred.
static DEBUG_A: Style = Yellow.foreground().bold();
static DEBUG_B: Style = Yellow.bold();

looks like we don't need the foreground call?

dcb9:
removed foreground call. 

`Style::new().fg(Yellow)` and `Yellow.foreground()` are equivalent, now I choose the first style, which one do you prefer?

DaniPopes (approved):
Thanks!

PR #108, opened 2024-04-19
fix: replace line endings on Windows to enforce deterministic metadata

klkvr:
I'm writing tests for https://github.com/foundry-rs/foundry/pull/7711 which are hardcoding CREATE2 library deployment addreses and they are currently failing on Windows probably because \r\n line endings which are causing metadata differences.

This PR tries to fix this by replacing such endings with \n


src/artifacts/mod.rs
```diff
@@ -1467,7 +1467,9 @@ impl Source {
     pub fn read(file: impl AsRef<Path>) -> Result<Self, SolcIoError> {
         let file = file.as_ref();
         trace!(file=%file.display());
-        let content = fs::read_to_string(file).map_err(|err| SolcIoError::new(err, file))?;
+        let content = fs::read_to_string(file)
+            .map_err(|err| SolcIoError::new(err, file))?
+            .replace("\r\n", "\n");
```
mattsse, line 1472: should we only do this on windows?

this is probably not a big perf hit, but if we can limit this to windows only, it would be ideal

mattsse (changes requested):
q about windows cfg.

let's also add a note for this

mattsse (approved):

PR #115, opened 2024-04-26
feat: compiler abstraction

klkvr:
This PR introduces following traits:
- `Compiler` - wrapper around compiler executable (`solc`/`vyper`) with following associated types
    - `Input` - trait capable of generating a `Vec<Self>` from given `Sources` set
    - `ParsedSource` - parser which extracts semver requirements and imported files from a given source
    - `CompilationError` - errors returned by underlying compiler binary
    - `Settings` - compiler settings
- `CompilerVersionManager` - factory of `Compiler`s capable of getting compilers by version

This PR contains 2 implementations of `Compiler`: `Solc` which is slightly updated `Solc` struct, and `Vyper` which is a draft PoC impl always invoking `vyper` command without any version management.

Currently it's already possible to compile and use cache for Vyper sources if `vyper` is avaialble on host machine.

Open questions:
1. Do we need `Compiler::Error` associated type? It requires additional generics on `Result`s returned by `ProjectCompiler` which I am not sure we'll ever use.
2. Right now all `Project::compile*` functions accept either `Compiler` or `CompilerVersionManager`. This makes API a bit more complex as earlier we'd just call `project.compile()` and it'd use svm to auto-detect compiler version. Perhaps it makes sense to add more helper fns/types?

Some parts of the code are ignored for now as they are not really used in the actual compiler pipeline. I'll work on making tests pass and adapting all code to changes.

Other general changes to types:
-  `allowed_paths` and `include_paths` are moved from `Project` to `ProjectPathsConfig`
- some of the compiler output types were indexing paths via `String`, updated keys for those to `PathBuf`.
- `Solc` now always required to keep a `Version` as we need it to configure all flags correctly.


src/compile/project.rs
```diff
@@ -102,37 +102,59 @@
...
+}
+
+impl<T, E> From<T> for MaybeCompilerError<E>
+where
+    T: Into<SolcError>,
+{
+    fn from(e: T) -> Self {
+        MaybeCompilerError::SolcError(e.into())
+    }
+}
```
mattsse, line 133: let's move to error.rs

src/artifact_output/mod.rs
```diff
@@ -1033,7 +1041,7 @@ impl ArtifactOutput for MinimalCombinedArtifacts {
 
     fn contract_to_artifact(
         &self,
-        _file: &str,
+        _file: &Path,
```
mattsse, line 1044: Path is better yes

src/compilers/mod.rs
mattsse, line 1: needs proper docs on all fns, types, traits

src/compilers/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+    type Input: CompilerInput<Settings = Self::Settings>;
+    type Error: CompilerError;
+    type CompilationError: CompilationError;
+    type ParsedSource: ParsedSource;
+    type Settings: CompilerSettings;
+
+    fn compile(
+        &self,
+        input: Self::Input,
+    ) -> Result<(Self::Input, CompilerOutput<Self::CompilationError>), Self::Error>;
```
mattsse, line 88: I assume we need to take ownership of the input?

mattsse (changes requested):
only skimmed, but this looks great, the compiler traits all make sense to me. 
I assume we'll change some of those eventually but this should do it for now.

left some nits but still need to take a closer look

maybe @DaniPopes also has some thoughts

src/compilers/mod.rs
klkvr: The idea here is that we might need to make changes to input (santize settings, strip paths), so we take ownership and return back modified version which was seen by compiler, so that we can then save it into build info

mattsse:
> Solc now always required to keep a Version as we need it to configure all flags correctly.

this makes sense, a solc that can't resolve the version will also fail when compiling, so `new(path) -> Result<Solc>` is more useful imo

src/compile/mod.rs
```diff
@@ -147,71 +149,21 @@ impl fmt::Display for SolcVersion {
...
-        self
-    }
-
-    /// Adds an argument to pass to the `solc` command.
-    #[must_use]
-    pub fn arg<T: Into<String>>(mut self, arg: T) -> Self {
-        self.args.push(arg.into());
-        self
+    /// A new instance which points to `solc`. Invokes `solc --version` to determine the version.
+    pub fn new(path: impl AsRef<Path>) -> Result<Self> {
```
mattsse, line 158: this makes sense, this guarantees that the solc binary is usable.

should mention that this throws an error if the binary does not exist

src/compile/mod.rs
```diff
@@ -505,106 +399,20 @@ impl Solc {
...
-    ///
-    /// ```no_run
-    /// use foundry_compilers::{CompilerInput, Solc};
-    ///
-    /// let solc = Solc::default();
-    /// let input = CompilerInput::new("./contracts")?[0].clone();
-    /// let output = solc.compile_exact(&input)?;
-    /// # Ok::<_, Box<dyn std::error::Error>>(())
-    /// ```
-    pub fn compile_exact(&self, input: &CompilerInput) -> Result<CompilerOutput> {
```
mattsse, line 535: can we keep this somehow,

still useful to invoke solc directly

src/compilers/mod.rs
mattsse: gotcha, this makes sense

mattsse (changes requested):
I also wonder how useful the separation of `T: Artifact` and `C: Compiler` still is

would it simplify things if we bake `T: Artifact` into the compiler abstraction? we could still have custom artifact handlers if we make this generic on the compiler.

for example:

```rust
struct SolcCompiler<T> {
  binary: Solc,
  handler: T
}
```

just an idea, maybe unfeasible

elfedy (commented):
Hey just wanted to mention that we are doing a [POC](https://github.com/Moonsong-Labs/compilers/pull/2) for implementing foundry compilers for zksync and this change would be a good step forward to set us up for success.
At first glance, the only roadblock for us implementing the `Compiler` trait is for it to also be generic over `Contract` as we would need a data structure with some [different fields](https://docs.zksync.io/zk-stack/components/compiler/toolchain/solidity.html#standard-json). Could give more info in a separate issue (or even propose something on a PR) once we have a working implementation and try implement the trait ourselves.
There's also the fact that we need to compile files to both EVM and zkEVM for a single `Project` instance, but that's probably something that could be discussed later.

src/compile/mod.rs
```diff
@@ -614,14 +392,50 @@ impl Solc {
     fn map_io_err(&self) -> impl FnOnce(std::io::Error) -> SolcError + '_ {
         move |err| SolcError::io(err, &self.solc)
     }
+
+    pub fn configure_cmd(&self) -> Command {
```
mattsse, line 396: docs would be nice here

mattsse (approved):
this is awesome,

vyper support is really nice!

one pedantic doc nit

PR #116, opened 2024-04-29
fix: add checks for != root folder

mattsse:
closes https://github.com/foundry-rs/foundry/issues/7811

prevents accidentally deleting the root folder


onbjerg (approved):
reasonable

PR #117, opened 2024-04-30
fix: short-circuit symlink cycle

mattsse:
closes https://github.com/foundry-rs/foundry/issues/7820

pnpm setups can contain symlink loops, we need to track visited symlink dirs and short circuit if we already visited a symlink dir

### DIFF
--- a/src/remappings.rs
+++ b/src/remappings.rs
@@ -674,11 +674,10 @@ fn find_remapping_candidates(
             // ```
             if entry.path_is_symlink() {
                 if let Ok(target) = utils::canonicalize(entry.path()) {
-                    if visited_symlink_dirs.contains(&target) {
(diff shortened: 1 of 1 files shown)


src/remappings.rs
```diff
@@ -665,6 +674,11 @@ fn find_remapping_candidates(
             // ```
             if entry.path_is_symlink() {
                 if let Ok(target) = utils::canonicalize(entry.path()) {
+                    if visited_symlink_dirs.contains(&target) {
+                        // short-circuiting if we've already visited the symlink
+                        return Vec::new();
+                    }
+                    visited_symlink_dirs.insert(target.clone());
```
DaniPopes, line 681: if !insert

PR #118, opened 2024-05-01
fix: enable all tests

klkvr:
it seems that we had incorrect `cargo nextest` args which caused tests under `./tests` to never run


mattsse (approved):

klkvr:
CI failure related to https://github.com/foundry-rs/foundry/issues/7797

Addressed this by rewriting path resolution logic for artifacts. Changes:
- We now just sort files by number of components when resolving paths instead of finding top-most
- We now keep track of a single `taken_paths_lowercase` instead of keeping contracts in groups keyed by `MappedContract`.
- Deduplicated logic for artifact paths creation for contracts and standalone sources by extracing most of the logic into `ArtifactOutput::get_artifact_path`

cc @mattsse

klkvr:
A lot of WIndows tests were outdated, updated those. Will address clippy failure in #115

@mattsse ptal, added some changes using `slash_paths` here and there

src/cache.rs
```diff
@@ -57,7 +57,14 @@ impl<S: CompilerSettings> CompilerCache<S> {
 
     /// Removes entry for the given file
     pub fn remove(&mut self, file: &Path) -> Option<CacheEntry<S>> {
-        self.files.remove(file)
+        if let Some(entry) = self.files.remove(file) {
```
DaniPopes, line 60: `.inspect(|entry| { ... })`

src/config.rs
```diff
@@ -67,6 +67,15 @@ impl ProjectPathsConfig {
         let graph = Graph::resolve_sources(self, sources)?;
         let ordered_deps = collect_ordered_deps(&flatten_target, self, &graph)?;
 
+        #[cfg(windows)]
+        let ordered_deps = {
+            use path_slash::PathBufExt;
+            ordered_deps
```
DaniPopes, line 73: `for p in &mut ordered_deps { *p = ... }`?
or since we only iterate it doesn't need collect

src/flatten.rs
```diff
@@ -191,6 +191,15 @@ impl Flattener {
 
         let ordered_sources = collect_ordered_deps(&target.to_path_buf(), &project.paths, &graph)?;
 
+        #[cfg(windows)]
+        let ordered_sources = {
+            use path_slash::PathBufExt;
+            ordered_sources
```
DaniPopes, line 197: same as above

DaniPopes (changes requested):
lgtm, pending clippy and @mattsse

src/cache.rs
klkvr: https://github.com/foundry-rs/compilers/actions/runs/8923860095/job/24508985313?pr=118#step:5:571

should I bump MSRV?

DaniPopes: Ah no then use `map`

klkvr:
Updated tests to use `fd_lock` for locking installer, should be less flaky now

mattsse (approved):
ty!

PR #120, opened 2024-05-07
feat: add a few Solc install helpers back

### LINKED ISSUE #119: Searching for the new equivalent of `Solc::detect_version(..)`
I'm looking for the new equivalent of `Solc::detect_version(..)` 

Additional Context
It was there in 0.3.0 but not in 0.4.0 which is breaking some changes for [aderyn's](https://github.com/Cyfrin/aderyn)

mattsse:
Closes #119

these got lost while moving more stuff to versionmanager, however the finding an installed versioned based on the requirement alone is still useful and should be preserved.

This can still be pushed to version manager, but we should keep it on Solc directly as well

### DIFF
--- a/src/compile/mod.rs
+++ b/src/compile/mod.rs
@@ -18,6 +18,7 @@ use std::{
 pub mod many;
 
 pub mod output;
+use crate::resolver::parse::SolData;
 pub use output::{contracts, info, sources};
 
 pub mod project;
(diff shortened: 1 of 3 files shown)


src/compile/mod.rs
```diff
@@ -18,6 +18,7 @@ use std::{
 pub mod many;
 
 pub mod output;
+use crate::resolver::parse::SolData;
```
DaniPopes, line 21: Wrong place

onbjerg (approved):

PR #122, opened 2024-05-10
fix: re-enable yul settings sanitization

klkvr:
Closes https://github.com/foundry-rs/foundry/issues/7869


mattsse (approved):
makes sense, omitting them is better than failing

PR #124, opened 2024-05-20
chore: swap generics on `Project`

klkvr:
Changes order of generics on `Project`. We'll be doing a lot of `Project<C>` and most of the time we use the same `ConfigurableArtifacts`


src/project_util/mod.rs
```diff
@@ -33,12 +33,15 @@ pub struct TempProject<T: ArtifactOutput = ConfigurableArtifacts, C: Compiler =
     /// temporary workspace root
```
DaniPopes, line 33: here too? TempProject

mattsse (approved):
lgtm

PR #125, opened 2024-05-21
feat: Vyper imports parser

klkvr:
We need to parse Vyper import directives for caching to work correctly.

This PR implements parser which is able to process imports of format `import a.b.c` and `from a.b.c import d`. Relative imports (`from ...a.b.c import d`) are also supported.

Tested on snekmate codebase.


mattsse (approved):
lgtm

PR #126, opened 2024-05-22
fix: relax trait bounds

klkvr:
Updates trait bounds to make it easier to work with `ProjectCompileOutput` on foundry side


DaniPopes (approved):

mattsse (approved):

mattsse:
we need to allow BSL-1.0 for the crate now in deny.toml

PR #128, opened 2024-05-27
feat: allow multiple languages for compilers

klkvr:
This changes `Compiler` trait to also have associated `Language` type which is supposed to be an enum with languages supported by the compiler. Also `CompilerVersionManager` trait is removed and compiler now should contain full logic for version resolution.

For version management we have to keep track of all different languages and versions to compile them with, thus I've rewritten some of the graph logic to use `HashMap<C::Language, HashMap<Version, Sources>>` during compilation.

Now, instead of representing a binary wrapper, `Compiler` is more like binary manager which resolves and installs binaries dynamically depending on requested `language` and `version`.

It's mostly ready, I just need to add logic for compilation with locked compiler version.


src/compilers/solc/mod.rs
```diff
@@ -1,39 +1,55 @@
...
+#[derive(Debug, Clone, Default)]
+#[non_exhaustive]
+pub struct SolcRegistry;
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[non_exhaustive]
+pub enum SolcLanguages {
+    Solidity,
+    Yul,
+}
```
mattsse, line 33: this needs some docs

src/compilers/solc/mod.rs
```diff
@@ -1,39 +1,55 @@
...
 };
 
-impl Compiler for Solc {
+#[derive(Debug, Clone, Default)]
+#[non_exhaustive]
+pub struct SolcRegistry;
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[non_exhaustive]
+pub enum SolcLanguages {
```
mattsse, line 30: nit: 
```suggestion
pub enum SolcLanguage {
```

src/compilers/mod.rs
```diff
@@ -136,45 +206,31 @@ impl<E> Default for CompilerOutput<E> {
     }
 }
 
+pub trait Language: Hash + Eq + Clone + Debug {
```
mattsse, line 209: should this also be Display and I think this should be `'static`

src/compilers/solc/mod.rs
```diff
@@ -44,85 +60,80 @@ impl Compiler for Solc {
...
+        all_versions
     }
+}
 
-    fn with_include_paths(mut self, include_paths: BTreeSet<PathBuf>) -> Self {
-        self.include_paths = include_paths;
-        self
-    }
+#[derive(Debug, Clone, Serialize)]
+pub struct SolcVerionedInput {
```
mattsse, line 84: typo

src/compilers/solc/mod.rs
```diff
@@ -1,39 +1,55 @@
...
+    Solidity,
+    Yul,
+}
+
+impl Language for SolcLanguages {
     const FILE_EXTENSIONS: &'static [&'static str] = SOLC_EXTENSIONS;
+}
 
-    type Input = SolcInput;
+impl Compiler for SolcRegistry {
```
mattsse, line 39: hmm, it's a bit weird that we implement compiler on registry, maybe we can rename this? so it's more obvious that this is compiler?

mattsse (changes requested):
this makes a lot of sense and simplifies things a lot.

I only have doc nits

src/compilers/solc/mod.rs
klkvr: yeah, not sure how to name it

perhaps `MultiSolc` or just `SolcCompiler`?

mattsse: I think SolcCompiler fits

DaniPopes: Sol Compiler Compiler

mattsse (approved):
gg

PR #130, opened 2024-06-01
feat: make CompactContractBytecodeCow implement Artifact

DaniPopes:


mattsse (approved):
lgtm

src/compilers/mod.rs
mattsse, line 1: this will likely conflict with #128

@klkvr should we exclude these changes here?

DaniPopes: it doesnt

DaniPopes: ah nvm, it's just formatting changes

PR #134, opened 2024-06-04
fix: Vyper 0.4 support

klkvr:
Vyper 0.4 has a separate source file entity - interfaces, those should be passed in JSON input under "interfaces" section and can't be present in output selection. This PR Addresses this.

Also I've introduced separate Vyper JSON artifacts due to source maps having different format in 0.4


src/compilers/vyper/output.rs
```diff
@@ -0,0 +1,159 @@
...
+impl VyperOutput {
+    /// Retains only those files the given iterator yields
+    ///
+    /// In other words, removes all contracts for files not included in the iterator
+    pub fn retain_files<'a, I>(&mut self, files: I)
+    where
+        I: IntoIterator<Item = &'a Path>,
+    {
+        // Note: use `to_lowercase` here because vyper not necessarily emits the exact file name,
+        // e.g. `src/utils/upgradeProxy.sol` is emitted as `src/utils/UpgradeProxy.sol`
```
mattsse, line 139: -.-

src/compilers/vyper/input.rs
```diff
@@ -22,17 +22,21 @@ pub struct VyperVersionedInput {
 }
 
 impl VyperInput {
-    pub fn new(sources: Sources, settings: VyperSettings) -> Self {
+    pub fn new(sources: Sources, mut settings: VyperSettings) -> Self {
         let mut new_sources = Sources::new();
         let mut interfaces = Sources::new();
 
         for (path, content) in sources {
             if path.extension().map_or(false, |ext| ext == "vyi") {
```
mattsse, line 30: nit: we could add constants for the extensions in VYPER_EXTENSIONS

src/compilers/vyper/output.rs
```diff
@@ -0,0 +1,159 @@
...
+use alloy_primitives::Bytes;
+use serde::Deserialize;
+use std::{
+    collections::{BTreeMap, HashSet},
+    path::{Path, PathBuf},
+};
+
+/// Before Vyper 0.4 source map was represented as a string, after 0.4 it is represented as a map
+/// where compressed source map is stored under `pc_pos_map_compressed` key.
+fn deserialize_sourcemap<'de, D>(deserializer: D) -> Result<Option<String>, D::Error>
```
mattsse, line 13: nit: move below type defs

mattsse, line 13:
```suggestion
fn deserialize_vyper_sourcemap<'de, D>(deserializer: D) -> Result<Option<String>, D::Error>
```

src/compilers/vyper/output.rs
```diff
@@ -0,0 +1,159 @@
...
+where
+    D: serde::Deserializer<'de>,
+{
+    #[derive(Deserialize)]
```
mattsse, line 17: should this be untagged?

src/compilers/vyper/output.rs
```diff
@@ -0,0 +1,159 @@
...
+
+#[derive(Clone, Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct VyperEvm {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub bytecode: Option<Bytecode>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub deployed_bytecode: Option<DeployedBytecode>,
+    /// The list of function hashes
+    #[serde(default, skip_serializing_if = "::std::collections::BTreeMap::is_empty")]
```
mattsse, line 77: the skip_serializing is redundant here because type is only Deserialize

mattsse (changes requested):
nice!

a few serde nits

src/compilers/vyper/output.rs
klkvr: This was copy-pasted and edited from solc impl, so not sure if that's actually true for Vyper :) 

those output impls will be deduped once we make artifacts generic

src/compilers/vyper/output.rs
klkvr: ah right, should've added tests right away

addressed in [2632f99](https://github.com/foundry-rs/compilers/pull/134/commits/2632f99f99789a7b3ef4ebf4126cd768daaee682)

mattsse (approved):

PR #135, opened 2024-06-05
refactor: sparse output

klkvr:
better approach for https://github.com/foundry-rs/foundry/pull/8061

We already keep all compilation settings on `Project` (paths, compier settings, compiler itself), so this PR moves `sparse_output` to `Project` as well to allow us to just configure it in `Config::project` instead of adding it to compiler on every invocation in multiple places.

Also this PR gets rid of separate `SparseOutputFileFilter<D>` trait and makes `sparse_output` filter just a `FileFilter` again. Instead, we now have `ParsedSource::compilation_dependencies` which determines which of the source imports must be compiled along with it.

Simplified some sparse output related logic along with changes.

Due to `Box<dyn FileFilter>` on `Project` this PR adds `dyn-clone` to keep `Clone` impl and `derivative` to avoid manually implementing `Debug`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,8 @@ itertools = "0.13"
 auto_impl = "1"
 
 winnow = "0.6"
+dyn-clone = "1"
+derivative = "2.2"
 
 [dev-dependencies]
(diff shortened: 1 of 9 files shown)


mattsse (approved):
nice

PR #136, opened 2024-06-05
fix: version resolution

klkvr:
ref https://github.com/foundry-rs/foundry/issues/8062
ref https://t.me/foundry_support/54144

introduced in https://github.com/foundry-rs/compilers/pull/128


mattsse (approved):
lgtm!

PR #139, opened 2024-06-07
fix: always fix windows line endings

klkvr:
Follow-up to https://github.com/foundry-rs/compilers/pull/108.

It seems that in certain cases files on WSL still have `\r\n` prefix, thus this PR changes logic to always fix Windows line endings and not only on Windows builds.

### DIFF
--- a/src/artifacts/mod.rs
+++ b/src/artifacts/mod.rs
@@ -1451,7 +1451,6 @@ impl Source {
         let content = fs::read_to_string(file).map_err(|err| SolcIoError::new(err, file))?;
 
         // Normalize line endings to ensure deterministic metadata.
-        #[cfg(windows)]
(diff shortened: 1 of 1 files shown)


src/artifacts/mod.rs
```diff
@@ -1532,7 +1531,6 @@ impl Source {
             tokio::fs::read_to_string(file).await.map_err(|err| SolcIoError::new(err, file))?;
 
         // Normalize line endings to ensure deterministic metadata.
-        #[cfg(windows)]
         let content = content.replace("\r\n", "\n");
```
DaniPopes, line 1534: can we wrap this with `if contains('\r')`

DaniPopes (approved):

Issue #129, opened 2024-05-31
Extract JSON artifacts to separate crates


DaniPopes:
Allows usage as a library without the build system stuff

PR #142, opened 2024-06-13
refactor: extract artifacts to a separate crate

klkvr:
Closes https://github.com/foundry-rs/compilers/issues/129

Divides `foundry-compilers` into 3 crates:
- `foundry-compilers-artifacts` - JSON artifacts for Solc and Vyper compilers
- `foundry-compilers-project` - Project related logic, including compiler abstraction and compilation pipeline
- `foundry-compilers-core` - error types and utilties shared by both crates


crates/artifacts/src/vyper/mod.rs
DaniPopes, line 1: can artifacts be split further into solc- and vyper- artifacts?

crates/artifacts/LICENSE-APACHE
DaniPopes, line 1: license files are not needed, `license.workspace = true` is enough

crates/artifacts/src/vyper/mod.rs
klkvr: yeah, but we're mostly operating on solc artifacts currently and just convert Vyper artifacts to them after deserialization, so vyper artifacts crate will be pretty small and will depend on solc artifacts

crates/project/src/lib.rs
```diff
@@ -42,39 +35,37 @@ pub use compile::{
...
+use cache::CompilerCache;
 use compile::output::contracts::VersionedContracts;
+use compilers::multi::MultiCompiler;
 use derivative::Derivative;
-use error::Result;
+use foundry_compilers_artifacts::{
+    output_selection::OutputSelection,
+    sources::{Source, Sources},
+    Contract, Settings, Severity, SourceFile, StandardJsonCompilerInput,
+};
```
mattsse, line 57: we likely want to re-export 

foundry_compilers_artifacts as artifacts

so that you don't need to depend on all crates if you're already pulling in compilers

mattsse (changes requested):
splitting into
artifacts
core
compilers

makes sense

I suggest to do a foundry companion first before we merge this so that we can fix things like missing re-exports etc.

crates/artifacts/src/vyper/mod.rs
DaniPopes: Understandable, but I would still like the separation for the structure and further additions

klkvr: makes sense! updated in https://github.com/foundry-rs/compilers/pull/142/commits/50e636cc3c36850dcd491e9f563203f2e459e0c7

crate name seems a bit too long at this point, should it be `foundry-artifacts-vyper` or `foundry-compilers-vyper`?

DaniPopes: it's fine

DaniPopes (changes requested):
last thing, artifacts should be a subdirectory like
- artifacts
  - solc/src -- base solc
  - vyper/src -- base vyper + imports solc
  - artifacts/src -- meta crate that re-exports both solc and vyper (might not be necessary cc @mattsse )

mattsse (approved):
I like this

image

PR #143, opened 2024-06-14
refactor(flatten): move compilation logic into `Flattener`

klkvr:
flattening requires specific compilation pipeline (AST-only, non-cached), it is an implementation detail which should be configured, run and tested on compilers level.


src/flatten.rs
```diff
@@ -173,25 +188,26 @@ pub struct Flattener {
...
-        let input_files = output
-            .artifacts_with_files()
-            .map(|(file, _, _)| PathBuf::from(file))
-            .collect::<HashSet<_>>()
-            .into_iter()
-            .collect::<Vec<_>>();
-
-        let sources = Source::read_all_files(input_files)?;
+        // Configure project to compile the target file and only request AST for target file.
+        let mut project = project.clone();
```
DaniPopes, line 200: take owned by input?

DaniPopes (approved):

DaniPopes (approved):

mattsse (approved):
nice

PR #157, opened 2024-06-28
chore: remove most impl AsRef<str,Path>

DaniPopes:
These mostly only bloat code size and compilation times, which are already not great when we are already codegenning a billion structs and ASTs with serde derives

For AsRef<str>, the only change is having to specify an extra 1 (one) character (&) when using `String` so it derefs to `str`, otherwise it's the same

For AsRef<Path> it's the same as AsRef<str> when using Path/PathBuf, but it's a bit worse for string literals; but most of these functions are called with variables in the first place. except in tests which will have to use .as_ref() or Path::new

I've not removed these when using with &[] or impl IntoIterator since that does help quite a bit, e.g if you have an owning iterator of Strings, you cannot borrow inside of a map closure

I've also kept Into<PathBuf> since that makes more sense, you want to avoid cloning if you can pass in an owned thing already, otherwise you'll clone it yourself

Also fixes some inconsistencies in function signatures for str vs path

### DIFF
--- a/crates/artifacts/solc/src/ast/lowfidelity.rs
+++ b/crates/artifacts/solc/src/ast/lowfidelity.rs
@@ -53,9 +53,9 @@ pub struct Node {
 
 impl Node {
     /// Deserialize a serialized node attribute.
-    pub fn attribute<D: DeserializeOwned>(&self, key: impl AsRef<str>) -> Option<D> {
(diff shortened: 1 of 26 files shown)


mattsse (approved):
reasonable, ty!

I assume we need to update a few calls in foundry after we release this

DaniPopes:
Pending github waking up and running actions

PR #159, opened 2024-06-28
feat: respect `paths.libraries` for Vyper

klkvr:
Vyper 0.4 added `search_paths` key which is basically equivalent to solc's `include_paths` which we are already populating during import resolution.

Adds this key to compiler input and also tries going through `paths.libs` while resolving Vyper imports.

This allows users to normally configure dependencies outside of their `src` directory (i.e. from `venv/.../site-packages` or `lib`)


mattsse (approved):

mattsse:
pedantic clippy

DaniPopes (approved):

PR #162, opened 2024-06-28
refactor: unify sources and filtered sources

DaniPopes:
Allows reusing resources and just doing `retain` instead of `.into_iter().collect()` on 3 nested maps


mattsse (approved):
nice

PR #166, opened 2024-07-10
feat: add Prague evm version

rakita:
Adds Prague spec.

TODO: PragueEOF support is needed.

### DIFF
--- a/crates/artifacts/solc/src/lib.rs
+++ b/crates/artifacts/solc/src/lib.rs
@@ -823,7 +823,6 @@ impl EvmVersion {
         if *version >= BYZANTIUM_SOLC {
             // If the Solc version is the latest, it supports all EVM versions.
             // For all other cases, cap at the at-the-time highest possible fork.
-
             let normalized = if *version >= PRAGUE_SOLC {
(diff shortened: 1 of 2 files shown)


mattsse (commented):
we can add this to the enum because solc already has support for this

https://github.com/ethereum/solidity/blob/5e9d38e9dc23d124b87bc3aa85714fd4db7c8820/liblangutil/EVMVersion.h#L63

but we can't add it to `normalize_version_solc` yet because we don't know the version yet and 0.8.26 is incorrect I believe because that's the latest release:

https://github.com/ethereum/solidity/releases/tag/v0.8.26

fgimenez:
> but we can't add it to `normalize_version_solc` yet because we don't know the version yet and 0.8.26 is incorrect I believe because that's the latest release:

good catch, I see in solidity's repo that the `prague` EVM version has been already introduced https://github.com/ethereum/solidity/pull/15152 not sure if this means it will be part of the next release?

mattsse (approved):
lgtm

DaniPopes (approved):

PR #169, opened 2024-07-16
fix: allow empty modifier body in AST

klkvr:
Closes https://github.com/foundry-rs/foundry/issues/7554

solc since 0.6.7 allows empty `body` on modifiers

this was incorrect in OpenZeppelin AST impl which served as a reference for our implementation https://github.com/OpenZeppelin/solidity-ast/pull/43


klkvr:
ah accidentally pushed by wip branch with smarter settings configuration

mattsse (approved):

PR #170, opened 2024-07-16
[wip] feat: allow multiple compiler configs

klkvr:
This adds `CompilerSettings::Restrictions` associated type which allows configuring constraints for compilation settings. Those constraints can be checked against a settings object and merged.

Ideally given a set of constraints we'd be able to automatically resolve configurations given only basic (default) config, though I think we can do this in follow-up. Current implementation requires user to provide us with concrete profiles such that for each defined constraint at least one of the profiles matches it. This should be enough for most of the usecases for now.

Profile to use is resolved in `Graph` similarly to version resolution.

The way this could look in foundry.toml is:
```toml

[profile.default]
# by default no via-ir and evm version is paris for everything
evm_version = "paris"
via_ir = false

# add cancun profile and via_ir profiles
additional_settings = [{ evm_version = "cancun" }, {via_ir = true} ]

# enforce compiling some contracts with cancun, and some with via-ir
# if those contracts are imported by some items in the project, constraints will apply as well
compilation_constraints = [{path = "./**/*.cancun.sol", min_evm_version = "cancun" }, {path = "SomeComplexFile.sol", via_ir = true }] 
```

This currently isn't really smart about which artifacts to write, so marking as WIP


crates/compilers/src/compilers/restrictions.rs
```diff
@@ -0,0 +1,45 @@
+use std::{
+    fmt::Debug,
+    ops::{Deref, DerefMut},
+};
+
+use semver::VersionReq;
+
+pub trait CompilerSettingsRestrictions: Copy + Debug + Sync + Send + Clone + Default {
+    fn merge(self, other: Self) -> Option<Self>;
+}
```
mattsse, line 10: needs docs

mattsse (commented):
all of this makes sense,
was able to follow, but could use some additional docs

mattsse (approved):

PR #171, opened 2024-07-16
feat: allow passing extra cli args to solc + some cleanup

klkvr:
This refactors `Compiler` a bit moving some of the configuration methods from `CompilerInput` to `CompilerSettings`, preferring sanitization of settings during input construction instead of in `with_*` methods of version-aware input.

Instead of using JSON solc settings, I've added separate `SolcSettings` type divided into JSON and CLI settings

### DIFF
--- a/crates/compilers/src/compilers/solc/mod.rs
+++ b/crates/compilers/src/compilers/solc/mod.rs
@@ -109,7 +109,7 @@ pub struct SolcVersionedInput {
     #[serde(flatten)]
     pub input: SolcInput,
     #[serde(flatten)]
-    cli_settings: CLISettings,
+    cli_settings: CliSettings,
 }
 
 impl CompilerInput for SolcVersionedInput {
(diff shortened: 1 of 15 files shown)


crates/compilers/src/compilers/solc/mod.rs
```diff
@@ -148,65 +144,106 @@ impl CompilerInput for SolcVersionedInput {
...
     }
+}
 
-    fn with_allow_paths(mut self, allowed_paths: BTreeSet<PathBuf>) -> Self {
-        self.allow_paths = allowed_paths;
-        self
-    }
+#[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct CLISettings {
```
mattsse, line 158:
```suggestion
pub struct CliSettings {
```

mattsse (approved):
lgtm, one naming suggestion

PR #174, opened 2024-07-17
feat: add `eofVersion` config option

klkvr:
adds key for enabling eof which is supported by https://github.com/ipsilon/solidity/tree/eof-functions-rebased

solc allows unknown keys in settings, so setting this wouldn't break default solc. though I think it might make sense to be under a feature flag. wdyt @mattsse

### DIFF
--- a/crates/artifacts/solc/Cargo.toml
+++ b/crates/artifacts/solc/Cargo.toml
@@ -30,6 +30,7 @@ tokio = { workspace = true, optional = true }
 tracing.workspace = true
 walkdir.workspace = true
 yansi.workspace = true
+serde_repr = "0.1"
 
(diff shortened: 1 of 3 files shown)


mattsse (approved):
makes sense but should we introduce an enum for this, like evmversion?

klkvr:
> makes sense but should we introduce an enum for this, like evmversion?

I believe it is a one-byte number by spec: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-3540.md#header

mattsse:
yeah could we make an enum for that?

klkvr:
Added enum in c04da87

mattsse (approved):

PR #176, opened 2024-07-26
fix: make remappings resolution more deterministic

klkvr:


crates/artifacts/solc/src/remappings.rs
```diff
@@ -226,13 +226,14 @@ impl Remapping {
...
         // all combined remappings from all subdirs
-        let mut all_remappings = HashMap::new();
+        let mut all_remappings = BTreeMap::new();
 
         let is_inside_node_modules = dir.ends_with("node_modules");
 
         let mut visited_symlink_dirs = HashSet::new();
         // iterate over all dirs that are children of the root
         for dir in walkdir::WalkDir::new(dir)
+            .sort_by_file_name()
```
mattsse, line 236: til

mattsse (approved):
makes sense

DaniPopes (approved):

PR #189, opened 2024-08-21
Incorrect Default EVM Version for Solidity Compiler 0.4.21-0.5.4

ZhangZhuoSJTU:
The default EVM version for Solidity compilers ranging from 0.4.21 to 0.5.4 is incorrectly implemented. Specifically, although the `Constantinople` hard fork had been introduced during this period, the default EVM version remains as `Byzantium`. After version 0.5.5, the default EVM version shifts to `Petersburg`. 

That is, `Constantinople` is never used as the default EVM version for Solidity compilers.

This can be confirmed by running `solc --help`, with the output indicating the default EVM version as follows:

```bash
$ solc --help
solc, the Solidity commandline compiler.

...

Allowed options:
  --help               Show help message and exit.
  --version            Show version and exit.
  --license            Show licensing information and exit.
  --evm-version version
                       Select desired EVM version. Either homestead,
                       tangerineWhistle, spuriousDragon, byzantium (default) or
                       constantinople.
  ...

$ solc --version
solc, the Solidity commandline interface
Version: 0.4.24+commit.e67f0147.Linux.g++
```

### DIFF
--- a/crates/artifacts/solc/src/lib.rs
+++ b/crates/artifacts/solc/src/lib.rs
@@ -816,6 +816,30 @@ pub enum EvmVersion {
 }
 
 impl EvmVersion {
+    /// Find the default EVM version for the given compiler version.
+    pub fn default_version_solc(version: &Version) -> Option<Self> {
+        // In most cases, Solc compilers use the highest EVM version available at the time.
+        let default = Self::default().normalize_version_solc(version)?;
+
+        // However, there are some exceptions where the default is lower than the highest available.
+        match default {
+            Self::Constantinople => {
+                // Actually, Constantinople is never used as the default EVM version by Solidity
+                // compilers.
+                Some(Self::Byzantium)
+            }
+            Self::Cancun if *version == Version::new(0, 8, 24) => {
+                // While Cancun is introduced at the time of releasing 0.8.24, it has not been
+                // supported by the mainnet. So, the default EVM version of Solc 0.8.24 remains as
+                // Shanghai.
+                //
+                // <https://soliditylang.org/blog/2024/01/26/solidity-0.8.24-release-announcement/>
+                Some(Self::Shanghai)
+            }
+            _ => Some(default),
+        }
+    }
+
     /// Normalizes this EVM version by checking against the given Solc [`Version`].
     pub fn normalize_version_solc(self, version: &Version) -> Option<Self> {
         // The EVM version flag was only added in 0.4.21; we work our way backwards
@@ -1900,6 +1924,40 @@ mod tests {
             serde_json::from_str::<SolcInput>(&pretty).unwrap_or_else(|err| {
                 panic!("Failed to read converted compiler input of {} {}", path.display(), err)
             });
+        }
+    }
+
+    #[test]
+    fn test_evm_version_default() {
+        for &(solc_version, expected) in &[
+            // Everything before 0.4.21 should always return None
+            ("0.4.20", None),
+            // Byzantium clipping
+            ("0.4.21", Some(EvmVersion::Byzantium)),
+            // Constantinople bug fix
+            ("0.4.22", Some(EvmVersion::Byzantium)),
+            // Petersburg
+            ("0.5.5", Some(EvmVersion::Petersburg)),
+            // Istanbul
+            ("0.5.14", Some(EvmVersion::Istanbul)),
+            // Berlin
+            ("0.8.5", Some(EvmVersion::Berlin)),
+            // London
+            ("0.8.7", Some(EvmVersion::London)),
+            // Paris
+            ("0.8.18", Some(EvmVersion::Paris)),
+            // Shanghai
+            ("0.8.20", Some(EvmVersion::Shanghai)),
+            // Cancun
+            ("0.8.24", Some(EvmVersion::Shanghai)),
+            ("0.8.25", Some(EvmVersion::Cancun)),
+        ] {
+            let version = Version::from_str(solc_version).unwrap();
+            assert_eq!(
+                EvmVersion::default_version_solc(&version),
+                expected,
+                "({version}, {expected:?})"
+            )
         }
     }
 


ZhangZhuoSJTU:
Additionally, in Solidity 0.8.24, `Cancun` is not set as the default (0.8.25 does so). [Reference](https://soliditylang.org/blog/2024/01/26/solidity-0.8.24-release-announcement/#:~:text=Thus%2C%20this%20new%20compiler%20release%20does%20not%20yet%20make%20it%20its%20default%20target)

crates/core/src/utils.rs
```diff
@@ -83,7 +79,7 @@ pub const SHANGHAI_SOLC: Version = Version::new(0, 8, 20);
 
 /// Cancun support
 /// <https://soliditylang.org/blog/2024/01/26/solidity-0.8.24-release-announcement/>
```
mattsse, line 81: can we add context here that this is made the default in 0.8.25?

mattsse (commented):
pending @klkvr

DaniPopes:
I dont think we were aiming for "solc's default" but rather the latest evm version supported at the given cersion?

ZhangZhuoSJTU:
> I dont think we were aiming for "solc's default" but rather the latest evm version supported at the given cersion?

@DaniPopes I think based on our usage of the `normalize_version_solc` function, it seems like it represents "solc's default." Is that correct, or am I misunderstanding something?

https://github.com/foundry-rs/block-explorers/blob/d44051a4e5fd4a522739b3c2be6df6aeebca849b/crates/block-explorers/src/contract.rs#L279-L281

```rust
    /// Returns the contract's compiler settings.
    #[cfg(feature = "foundry-compilers")]
    pub fn settings(&self) -> Result<Settings> {
        let mut settings = self.source_code.settings()?.unwrap_or_default();

        if self.optimization_used == 1 && !settings.optimizer.enabled.unwrap_or_default() {
            settings.optimizer.enable();
            settings.optimizer.runs(self.runs as usize);
        }

        settings.evm_version = self.evm_version()?;

        Ok(settings)
    }
    
    ...

    /// Parses the EVM version.
    #[cfg(feature = "foundry-compilers")]
    pub fn evm_version(&self) -> Result<Option<EvmVersion>> {
        match self.evm_version.to_lowercase().as_str() {
            "" | "default" => {
                Ok(EvmVersion::default().normalize_version_solc(&self.compiler_version()?))
            }
            _ => {
                let evm_version = self
                    .evm_version
                    .parse()
                    .map_err(|e| EtherscanError::Unknown(format!("bad evm version: {e}")))?;
                Ok(Some(evm_version))
            }
        }
    }
```

https://github.com/foundry-rs/compilers/blob/3f080851e3a38a2a3e867f92173d85449aa07fb4/crates/artifacts/solc/src/lib.rs#L820-L852

DaniPopes:
the comment in the function itself suggests otherwise, "cap at the highest possible fork"

basically if a user specifies cancun, but the solc version only knows up to shanghai, then we'll cap to shanghai

ZhangZhuoSJTU:
@DaniPopes, thank you for your prompt response! I apologize if I'm misunderstanding something, but I have a couple of questions for clarification:

1) In the `block-explorer` [repo](https://github.com/foundry-rs/block-explorers/blob/d44051a4e5fd4a522739b3c2be6df6aeebca849b/crates/block-explorers/src/contract.rs#L253-L264), it seems the purpose of the normalization process is to determine the EVM version when the configuaration specifies it as "default" or even does not provided. To this end, finding the "solc's default" EVM version appears particularly useful when building a local or project from Etherscan. Did I get that right? It seems like this might be a more common use case (comparing to "capping at the highest possible fork").

2) Additionally, could you please provide any concrete use cases for "capping at the highest possible fork"? Is there any code reference available that illustrates this? According to the [solidity documentation](https://soliditylang.org/blog/2024/01/26/solidity-0.8.24-release-announcement/#:~:text=Thus%2C%20this%20new%20compiler%20release%20does%20not%20yet%20make%20it%20its%20default%20target), it is not always a good idea to use a version that is "not actually deployed to the Ethereum mainnet" but the highest possible fork.

ZhangZhuoSJTU:
Or, maybe it is the block-explorer [repo](https://github.com/foundry-rs/block-explorers/blob/d44051a4e5fd4a522739b3c2be6df6aeebca849b/crates/block-explorers/src/contract.rs#L253-L264) misuses the API. 

In that case, I guess a proper patch is to introduce a new API in the `compilers` repo, and then change the `block-explorer` repo.

klkvr:
> 1. In the `block-explorer` [repo](https://github.com/foundry-rs/block-explorers/blob/d44051a4e5fd4a522739b3c2be6df6aeebca849b/crates/block-explorers/src/contract.rs#L253-L264), it seems the purpose of the normalization process is to determine the EVM version when the configuaration specifies it as "default" or even does not provided. To this end, finding the "solc's default" EVM version appears particularly useful when building a local or project from Etherscan. Did I get that right? It seems like this might be a more common use case (comparing to "capping at the highest possible fork").

Yeah, this logic in block-explorers is not great, we should probably keep a similar mapping from solc version to default EVM version for it.

> 2. Additionally, could you please provide any concrete use cases for "capping at the highest possible fork"? Is there any code reference available that illustrates this? According to the [solidity documentation](https://soliditylang.org/blog/2024/01/26/solidity-0.8.24-release-announcement/#:~:text=Thus%2C%20this%20new%20compiler%20release%20does%20not%20yet%20make%20it%20its%20default%20target), it is not always a good idea to use a version that is "not actually deployed to the Ethereum mainnet" but the highest possible fork.

Primary purpose of this is to support multi-version projects. E.g. if some of user contracts are getting compiled with 0.8.25, and some with 0.8.18, and he has `evm_version = "cancun"` in foundry.toml, then 0.8.25 would get called with "cancun" evm version, and for 0.8.18 we will normalize it to "paris"

ZhangZhuoSJTU:
@DaniPopes @klkvr I've updated a new patch, where we introduce a new function named `default_version_solc` to support the functionality of finding the default version.

If it is a proper patch, I will also change the code in `block-explorers` accordingly.

Thanks again!

ZhangZhuoSJTU:
@DaniPopes @klkvr @mattsse kindly pinging for your review when you have a moment. Thanks!

klkvr (approved):
this is a bit hacky, but still works better than current approach for determining default EVM version.

if this turns out not great, we can always add a similar mapping from solc version to default EVM version

mattsse (approved):

PR #192, opened 2024-08-30
feat: add `snapshots` directory

zerosnacks:
This directory holds gas snapshots that users would usually want to check into their git

Required for https://github.com/foundry-rs/foundry/pull/8755 (https://github.com/foundry-rs/foundry/issues/2065)


klkvr:
hmm, I think we can keep this on foundry's `Config`? given that it's just an internal cheatcode configuration

zerosnacks:
> hmm, I think we can keep this on foundry's `Config`? given that it's just an internal cheatcode configuration unrelated to compilation

Fair point, moving to draft for now - seeing if I can implement this in an elegant way directly in Foundry

zerosnacks:
Added in directly in Foundry, closing

Issue #43, opened 2024-01-12
Improve incompatible solidity versions error formatting


mattsse:
ref https://github.com/foundry-rs/foundry/issues/6722

this should improve error formatting and ideally highlight the conflicting versions

https://github.com/foundry-rs/compilers/blob/314c9171c3c09a90dd528c4884bc1b3da569308b/src/resolver/mod.rs#L607-L609

AbhinavMir:
would love to take this on sometime this week. :)

AbhinavMir:
@mattsse How do I get `foundry_compilers` to throw errors when compiling faulty code? I wrote a Solidity code with syntax errors, and expect it to fail, but `foundry_compilers` compiles it.

```rust
use foundry_compilers::{Project, ProjectPathsConfig};

fn main() {
    let project = Project::builder()
        .paths(ProjectPathsConfig::hardhat("/home/august/bad_code_solidity_hardhat").unwrap())
        .build()
        .unwrap();

    if let Err(e) = project.compile() {
        // The compilation failed, `e` contains the error.
        eprintln!("Compilation failed with error: {:?}", e);
    }

    project.rerun_if_sources_changed();
}
```

PR #200, opened 2024-09-16
feat: better error messages for incompatible versions

klkvr:
Closes https://github.com/foundry-rs/compilers/issues/43

Adds logic to resolve which nodes actually caused the resolver failure. This provides us access to ids of actually incompatible nodes in `format_imports_list` fn. 

It will now highlight sources causing failure in red

### DIFF
--- a/crates/compilers/src/resolver/mod.rs
+++ b/crates/compilers/src/resolver/mod.rs
@@ -547,6 +547,7 @@ impl<L: Language, D: ParsedSource<Language = L>> Graph<D> {
     fn format_imports_list<W: std::fmt::Write>(
         &self,
         idx: usize,
+        imports: impl IntoIterator<Item = usize>,
         f: &mut W,
     ) -> std::result::Result<(), std::fmt::Error> {
         let node = self.node(idx);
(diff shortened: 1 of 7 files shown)


crates/compilers/src/resolver/mod.rs
```diff
@@ -547,37 +548,92 @@ impl<L: Language, D: ParsedSource<Language = L>> Graph<D> {
...
+                    self.format_imports_list(idx, [*node, failed_node_idx].into(), &mut msg)
+                        .unwrap();
+                    return Err(msg);
+                }
             }
         }
+
+        let mut msg = "Found incompatible versions:\n".white().to_string();
+        self.format_imports_list(idx, nodes.into_iter().collect(), &mut msg).unwrap();
+        Err(format!("Found incompatible versions:\n{msg}"))
```
DaniPopes, line 636: this string is repeated twice

DaniPopes (commented):
Is there a way we can add a test with snapbox, at least for the formatting?

DaniPopes (approved):

PR #208, opened 2024-10-01
feat: better extra_args handling

klkvr:
ref https://github.com/foundry-rs/foundry/issues/8997

Adds `Solc` constructor for creating solc instances with arguments. Changes `extra_args` to being prepended to actual solc arguments instead of appended as for now.

This allows us to use eof binary through `Solc::new_with_args("docker", ["run", ... ])`

We didn't end up using those for `--eof-version`, so this shouldn't break anything. Will add a test for this in scope `--eof` flag on foundry side

### DIFF
--- a/crates/compilers/src/compilers/solc/compiler.rs
+++ b/crates/compilers/src/compilers/solc/compiler.rs
@@ -447,7 +447,7 @@ impl Solc {
     /// Invokes `solc --version` and parses the output as a SemVer [`Version`].
     #[instrument(level = "debug", skip_all)]
     pub fn version(solc: impl Into<PathBuf>) -> Result<Version> {
-        Self::version_with_args(solc, &Vec::new())
+        Self::version_with_args(solc, &[])
     }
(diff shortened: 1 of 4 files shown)


crates/compilers/src/compilers/solc/compiler.rs
```diff
@@ -428,10 +446,11 @@ impl Solc {
 
     /// Invokes `solc --version` and parses the output as a SemVer [`Version`].
     #[instrument(level = "debug", skip_all)]
-    pub fn version(solc: impl Into<PathBuf>) -> Result<Version> {
-        crate::cache_version(solc.into(), |solc| {
+    pub fn version(solc: impl Into<PathBuf>, args: &Vec<String>) -> Result<Version> {
```
mattsse, line 449: we could keep the function as is but introduce a new one that accepts a slice

mattsse (approved):
lgtm, one suggestion

PR #213, opened 2024-10-16
fix: correctly handle b as pre-release in Vyper version

klkvr:
Closes https://github.com/foundry-rs/foundry/issues/9123

I've done similar workaround for rc before. For semver we need those to be prepended by a dash.

PEP440 pragmas are still might not be parsed correctly but this is not a showstopper as we're just ignoring them right now, and user would see a vyper error in case of mismatch.


DaniPopes:
there's also "a"

PR #216, opened 2024-11-18
fix: sanitize `settings.optimizer.details.inliner`

grandizzy:
Ref https://github.com/foundry-rs/foundry/issues/9322#issuecomment-2480484349
https://github.com/ethereum/solidity/issues/15576#issuecomment-2480385494

Set inliner to None since it is properly supported from 0.8.5 version( see `Standard JSON: Properly allow the inliner setting under settings.optimizer.details` https://soliditylang.org/blog/2021/06/10/solidity-0.8.5-release-announcement/)


DaniPopes (approved):

DaniPopes (changes requested):
Actually can we put this in `sanitize`? And call that in Foundry

DaniPopes (approved):

PR #224, opened 2024-11-19
refactor: make Contract generic for Compiler and add metadata to CompilerOutput

elfedy:
Right now the compiler abstraction has two couplings that force foundry-zksync to implement it's [own fork of compilers](https://github.com/Moonsong-Labs/compilers/tree/zksync-v0.11.6):
1. `Contract` is specific to `solc`/EVM contracts. Era VM contracts, while requiring different fields, still could use most of the functionality of the `compilers` pipeline.
2. `CompilerOutput` has `solc` specific fields. `zksolc` compilation has relevant information that is useful to have later on, for example when storing `BuildInfo`

This PR implements changes to address this. If implemented, it would allow `foundry-zksync` (and potentially other non EVM implementations of foundry) to get rid of all overrides and only maintain ZKsync specific data structures/trait implementations. See [sample PR](https://github.com/Moonsong-Labs/compilers/pull/42). Changes include:
1. Make `Compiler` generic over `Contract` (using `CompilerContract` as a trait type).
2. Add `metadata` field to `CompilerOutput` in order to add arbitrary data to compilation output.


mattsse (changes requested):
I see, imo this seems reasonable if this simplifies things for zksync a lot.

the additional type/generic isn't too bad

left some doc nits, 
needs another review by @klkvr 

It'd be great if you could open a pr to foundry where the compilers deps a patched to this branch so we can ensure that this still works or what we need to tune

crates/compilers/src/compilers/mod.rs
```diff
@@ -201,16 +202,18 @@ pub trait CompilationError:
...
     #[serde(default = "Vec::new", skip_serializing_if = "Vec::is_empty")]
     pub errors: Vec<E>,
-    #[serde(default)]
-    pub contracts: FileToContractsMap<Contract>,
+    #[serde(default = "BTreeMap::new")]
+    pub contracts: FileToContractsMap<C>,
     #[serde(default)]
     pub sources: BTreeMap<PathBuf, SourceFile>,
+    #[serde(default)]
+    pub metadata: BTreeMap<String, String>,
```
mattsse, line 213: are these really just strings or are those json objects?

mattsse, line 213: I'd like to make this an option and skip if missing because now this will always be present for solc

crates/compilers/src/compile/output/contracts.rs
```diff
@@ -11,11 +11,25 @@ use std::{
...
+);
+
+impl<C> Default for VersionedContracts<C>
+where
+    C: CompilerContract,
+{
+    fn default() -> Self {
+        Self(BTreeMap::new())
+    }
+}
```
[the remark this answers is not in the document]
elfedy: removed the `CompilerContract` bound: https://github.com/foundry-rs/compilers/pull/224/commits/2be52cf83b50e1fe19b101aba19f9c4aec0627da

crates/compilers/src/compilers/mod.rs
```diff
@@ -267,6 +276,42 @@ pub trait Language:
     const FILE_EXTENSIONS: &'static [&'static str];
 }
 
+pub trait CompilerContract: Serialize + Send + Sync + Debug + Clone + Eq + Sized {
```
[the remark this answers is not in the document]
elfedy: Added, took them from original `Contract`: https://github.com/foundry-rs/compilers/pull/224/commits/0a36f5798f8bb92dea9acacac3cef072b2c665ae

crates/compilers/src/compilers/mod.rs
elfedy: Changed value to json: https://github.com/foundry-rs/compilers/pull/224/commits/c2a6c688ae8048c3b02db8f3268c72017893000f

elfedy: Skipped serializing if empty: https://github.com/foundry-rs/compilers/pull/224/commits/c2a6c688ae8048c3b02db8f3268c72017893000f is that good enough?

elfedy:
@mattsse thanks, added draft PR on foundry that compiles: https://github.com/foundry-rs/foundry/pull/9423

crates/compilers/src/compilers/mod.rs
```diff
@@ -267,6 +276,43 @@ pub trait Language:
     const FILE_EXTENSIONS: &'static [&'static str];
 }
 
+/// Represents a compiled contract
+pub trait CompilerContract: Serialize + Send + Sync + Debug + Clone + Eq + Sized {
+    fn abi_ref(&self) -> Option<&JsonAbi>;
+    fn bin_ref(&self) -> Option<&BytecodeObject>;
+    fn bin_runtime_ref(&self) -> Option<&BytecodeObject>;
```
klkvr, line 283:
```suggestion
    /// Reference to contract ABI
    fn abi_ref(&self) -> Option<&JsonAbi>;
    
    //// Reference to contract bytecode
    fn bin_ref(&self) -> Option<&BytecodeObject>;
    
    //// Reference to contract runtime bytecode
    fn bin_runtime_ref(&self) -> Option<&BytecodeObject>;
```

klkvr (changes requested):
very nice! thanks for working on this @elfedy

just have a small nit

I think it might also be useful to abstract the source eventually and have something like 
```rust
trait CompilerOutput {
     type Error;
     type Contract;
     type Source;
}
```

but this can definitely be refactored later once we have a nice way to support generic contracts in foundry

klkvr (approved):
lgtm!

mattsse (approved):

PR #239, opened 2025-01-07
feat: add more features to reduce dependencies

DaniPopes:
When importing sub crates directly (like artifacts) we don't need regex, walkdir, and other deps.

No functional changes intended.

### DIFF
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -18,25 +18,37 @@ workspace = true
 alloy-primitives.workspace = true
 cfg-if.workspace = true
 dunce.workspace = true
-path-slash.workspace = true
-regex.workspace = true
 semver.workspace = true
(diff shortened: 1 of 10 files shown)


crates/core/Cargo.toml
```diff
@@ -18,25 +18,37 @@ workspace = true
...
 [dev-dependencies]
 tempfile.workspace = true
 
 [features]
+async = ["dep:tokio"]
 project-util = ["dep:tempfile", "dep:fs_extra"]
+regex = ["dep:regex"]
 svm-solc = ["dep:svm", "dep:tokio"]
-async = ["dep:tokio"]
+walkdir = ["dep:walkdir"]
```
mattsse, line 52: makes sense

mattsse (approved):

PR #247, opened 2025-02-06
fix: ordering for flattener

klkvr:
Closes https://github.com/foundry-rs/foundry/issues/9788

ref https://github.com/foundry-rs/foundry/issues/9788#issuecomment-2633799377

Instead of sorting by path, firstly attempts sorting by filenames

### DIFF
--- a/crates/compilers/src/flatten.rs
+++ b/crates/compilers/src/flatten.rs
@@ -854,7 +854,24 @@ pub fn collect_ordered_deps<D: ParsedSource + MaybeSolData>(
         paths_with_deps_count.push((path_deps.len(), path));
     }
 
-    paths_with_deps_count.sort();
+    paths_with_deps_count.sort_by(|(count_0, path_0), (count_1, path_1)| {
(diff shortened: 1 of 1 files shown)


crates/compilers/src/flatten.rs
```diff
@@ -854,7 +854,24 @@ pub fn collect_ordered_deps<D: ParsedSource + MaybeSolData>(
         paths_with_deps_count.push((path_deps.len(), path));
     }
 
-    paths_with_deps_count.sort();
+    paths_with_deps_count.sort_by(|(count_0, path_0), (count_1, path_1)| {
+        // Compare dependency counts
+        match count_0.cmp(count_1) {
+            o if !o.is_eq() => return o,
+            _ => {}
+        };
```
mattsse, line 862: ah this isn't quite the same as 

```rust
if count_0 != count_1 {
  return 
}
```

because this handles all other variants

mattsse (approved):
makes sense lgtm

PR #258, opened 2025-03-25
chore: update file extension for compatibility

iamslown:
Hi! I updated the file extension to ensure compatibility with the system requirements. This change maintains functionality while following the expected file format standards.


mattsse (approved):
I let this one slide, because this actually fixes a deadlink

PR #268, opened 2025-05-16
fix: do not check if artifact files are dirty

grandizzy:


crates/compilers/src/cache.rs
```diff
@@ -989,7 +989,9 @@ impl<T: ArtifactOutput<CompilerContract = C::CompilerContract>, C: Compiler>
         for artifacts in self.cached_artifacts.values() {
             for artifacts in artifacts.values() {
                 for artifact_file in artifacts {
-                    if self.project.artifacts_handler().is_dirty(artifact_file).unwrap_or(true) {
+                    if !artifact_file.file.starts_with(&self.project.paths.artifacts)
```
mattsse, line 992: what's an example for an artifact_file that doesn't reside in the artifacts folder?

mattsse (changes requested):
I think this makes sense but the extra `!` is a bit confusing, can we add a note here?

crates/compilers/src/cache.rs
grandizzy: thanks for pushing back, I got this totally wrong. the reason here is that we have `additional_values` and `additional_files` and we didn't account the additional files when comparing checking if artifact dirty in cache, changed with https://github.com/foundry-rs/compilers/pull/268/commits/abb1371226c232ce6ec5f8e9b844f2a6d0b47e1e

PR #269, opened 2025-05-21
Support `transient` in `StorageLocation`

wtdcode:
Trivial patch. The new variant `transient` seems missing.


mattsse (approved):
lgtm, ty!

PR #271, opened 2025-05-23
Some fields are optional during `"stopAfter":"parsing"`

wtdcode:
Trivial patch.

Reproduce:

```bash
echo '{"language":"Solidity","sources":{"test.sol":{"content":"contract TT {\\n    function main() public {}\\n}"}},"settings":{"stopAfter":"parsing","optimizer":{"enabled":false,"runs":200},"outputSelection":{"*":{"":["ast"],"*":[]}},"viaIR":true,"libraries":{}}}' | solc --standard-json
```

Since there is no semantics analysis when stopping after parsing, `scope` along with a few other fields are not available and this PR marks all `scope` to be optional.

### DIFF
--- a/crates/artifacts/solc/src/ast/mod.rs
+++ b/crates/artifacts/solc/src/ast/mod.rs
@@ -178,7 +178,7 @@ ast_node!(
         fully_implemented: bool,
         linearized_base_contracts: Vec<usize>,
         nodes: Vec<ContractDefinitionPart>,
-        scope: usize,
+        scope: Option<usize>,
         #[serde(default, deserialize_with = "serde_helpers::default_for_null")]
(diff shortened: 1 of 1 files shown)


mattsse (approved):
makes sense,

do you need another patch for this right away?

wtdcode:
> makes sense,
> 
> do you need another patch for this right away?

Thanks for the instant response. I'm looking into solidity code to check all fields not emitted in this case.

Will raise another PR when I find more fields.

wtdcode:
Pretty hard to determine the exact full set of fields that need to be optional from the source code since they do not have schema like `serde`. Will get another PR when I finish parsing more solidity code =).

PR #272, opened 2025-05-26
draft: switch to prague by default

zerosnacks:


crates/compilers/Cargo.toml
```diff
@@ -62,7 +62,7 @@ tracing-subscriber = { version = "0.3", default-features = false, features = [
 similar-asserts.workspace = true
 fd-lock = "4.0.0"
 tokio = { version = "1.35", features = ["rt-multi-thread", "macros"] }
-reqwest = "0.12"
+reqwest = "=0.12.15"
```
zerosnacks, line 65: Note: https://github.com/seanmonstar/reqwest/issues/2699

`>= 0.12.16` currently fails to compile with `--no-default-features`

PR #273, opened 2025-05-26
chore: clean up error! calls

DaniPopes:
I got a `ERROR` log when using `forge flatten` with no extra message, and the error itself gets ignored with a fallback


mattsse (approved):
makes sense

PR #274, opened 2025-05-27
`resolc` version parsing update

pkhry:
### Description 

Updates version parsing from `resolc` stdout `--version` command to account for v1.* releases


pkhry:
accident, sorry

PR #292, opened 2025-07-23
Remove duplicate assembly check in is_dirty

Galoretka:
This commit removes a redundant check for assembly in the is_dirty method, leaving only a single, correct check for assembly. No other logic was changed.


mattsse (approved):
this is indeed duplicated

PR #301, opened 2025-08-21
chore: use svm instead of manual svm dir logic

DaniPopes:
I don't really know why these re-implement svm functions

Closes https://github.com/foundry-rs/compilers/pull/296


0xrusowsky (approved):
lgtm

PR #312, opened 2025-08-29
fix: make sources' paths absolute

DaniPopes:
Cache entries are keyed by absolute path relative to project root, but it's possible to construct and pass in Sources from a relative path, such as CLI argument in `forge lint`/`forge eip712`...

In these cases the path will be relative and will always be a cache miss. This includes the path's imports, which get resolved as relative to the initial path rather than root now with Solar.

Make all paths in Sources absolute when constructing Graph so that this change propagates through the entire compilation process.

I couldn't find a reproducer in today's stable/nightly Foundry, however making `config.solar_project()` use `project()` instead of `ephemeral_project()` will reproduce the behavior explained above.

With this change, `solar_project` will almost never call to solc.

### DIFF
--- a/crates/compilers/src/resolver/mod.rs
+++ b/crates/compilers/src/resolver/mod.rs
@@ -400,6 +400,9 @@ impl<P: SourceParser> Graph<P> {
             Ok(())
         }
 
+        // The cache relies on the absolute paths relative to the project root as cache keys.
+        sources.make_absolute(&paths.root);
+
         let mut parser = P::new(paths.with_language_ref());
 
         // we start off by reading all input files, which includes all solidity files from the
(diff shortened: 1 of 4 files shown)


crates/compilers/src/resolver/mod.rs
```diff
@@ -400,6 +400,9 @@ impl<P: SourceParser> Graph<P> {
             Ok(())
         }
 
+        // The cache relies on the absolute paths relative to the project root as cache keys.
+        sources.make_absolute(&paths.root);
```
mattsse, line 404: ah I believe that makes sense

can't think of any issues this cause

mattsse (approved):

PR #314, opened 2025-09-03
perf: parallelize Remapping::get_many

DaniPopes:
No logic changes except: 2nd commit removes walkdir and doesn't sort by file name anymore because order doesn't matter when we run it in parallel

### DIFF
--- a/crates/artifacts/solc/Cargo.toml
+++ b/crates/artifacts/solc/Cargo.toml
@@ -31,9 +31,6 @@ regex.workspace = true
 tokio = { workspace = true, optional = true, features = ["fs"] }
 futures-util = { workspace = true, optional = true }
 
-# walkdir
-walkdir = { workspace = true, optional = true }
-
 # rayon
 rayon = { workspace = true, optional = true }
 
@@ -48,5 +45,5 @@ foundry-compilers-core = { workspace = true, features = ["test-utils"] }
 [features]
 async = ["dep:tokio", "dep:futures-util"]
 checksum = ["foundry-compilers-core/hasher"]
-walkdir = ["dep:walkdir", "foundry-compilers-core/walkdir"]
+walkdir = ["foundry-compilers-core/walkdir"]
 rayon = ["dep:rayon"]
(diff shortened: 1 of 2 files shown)


crates/artifacts/solc/Cargo.toml
```diff
@@ -48,5 +45,5 @@ foundry-compilers-core = { workspace = true, features = ["test-utils"] }
 [features]
 async = ["dep:tokio", "dep:futures-util"]
 checksum = ["foundry-compilers-core/hasher"]
-walkdir = ["dep:walkdir", "foundry-compilers-core/walkdir"]
+walkdir = ["rayon", "foundry-compilers-core/walkdir"]
```
DaniPopes, line 48: `find` module and related functions are gated by "walkdir", don't wanna change feature gate

crates/artifacts/solc/src/remappings/find.rs
```diff
@@ -339,9 +320,9 @@ fn find_remapping_candidates(
             // ├── dep/node_modules
             //     ├── symlink to `my-package`
             // ```
-            if entry.path_is_symlink() {
-                if let Ok(target) = utils::canonicalize(entry.path()) {
-                    if !visited_symlink_dirs.insert(target.clone()) {
+            if file_type.is_symlink() {
```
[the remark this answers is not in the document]
DaniPopes: fixed here https://github.com/foundry-rs/compilers/pull/314/commits/8dfa9261ba1494e5d32d6cb410d7d20ab570804e

crates/artifacts/solc/Cargo.toml
mattsse: makes sense

crates/artifacts/solc/src/remappings/find.rs
```diff
@@ -76,41 +80,30 @@ impl Remapping {
...
-        let mut all_remappings = BTreeMap::new();
-
         let is_inside_node_modules = dir.ends_with("node_modules");
+        let visited_symlink_dirs = Mutex::new(HashSet::new());
 
-        let mut visited_symlink_dirs = HashSet::new();
         // iterate over all dirs that are children of the root
-        for dir in walkdir::WalkDir::new(dir)
-            .sort_by_file_name()
-            .follow_links(true)
```
mattsse, line 88: we dont need to follow symlinks explicitly here, right?

PR #318, opened 2025-09-15
merge foundry compilers v0.18.2

alexggh:
Merge foundry compilers v0.18.2 from upstream this is the version used by foundry stable release `v1.3.4`


alexggh:
Wrong repo.

PR #320, opened 2025-09-15
chore(`ci`): harden workflow by setting default permission to read only

zerosnacks:
Defines per action permissions scoping, defaulting to read only

RE: 

```
      - uses: actions/checkout@v5
        with:
          persist-credentials: false
```

See: https://github.com/actions/checkout/issues/485

This does not yet introduce pinning to hashes, I want to find a good way to maintain this first. Apparently Dependabot has a feature for this.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,48 @@
+name: CodeQL
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    branches: ["main"]
+  schedule:
+    - cron: "0 0 * * *"
+  workflow_dispatch:
+
+concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -72,11 +87,13 @@ jobs:
...
     timeout-minutes: 30
+    permissions:
+      actions: write
     steps:
       - uses: actions/checkout@v5
-      - uses: dtolnay/rust-toolchain@stable
         with:
-          components: clippy
+          persist-credentials: false
+      - uses: dtolnay/rust-toolchain@clippy
```
zerosnacks, line 96: note: this now uses `@clippy` in line with other Foundry crates which is `nightly + clippy` shorthand

PR #354, opened 2026-01-22
fix: serialize sources in build info ordered by source unit ID

### LINKED ISSUE #178: `sources` key in build info is no longer in order of source unit id
The build-info file used to have the same ordering as the solc JSON output and the index of a source path in the `sources` key would be the same as it's source unit id e.g. the 152nd index in `sources` list would be the source path of the source unit with an id of 152. Having the exact output of the compiler is important to avoid having to special case build frameworks. I think one of the changes having to do with caching or incremental compilation has broken this.

It is unclear what is the expected behavior in Foundry build artifacts and whether source id's will be duplicated or internally consistent e.g. in source maps, source offsets, etc. How does this work when cached artifacts are used? It would be great to have this behavior documented and maybe have more discussion on whether caching strategy's already do or, in the future, could provide fidelity to solc and Hardhat's build format.

As an aside, it would be nice to have more testing for the build-info as it used to be equivalent to Hardhat and has had subtle breaking changes in the last couple months. When I look at PR's touching this code, the changes are not consistently tested nor do fixes consistently have regression tests.

gakonst:
Fixes #178

## Summary

The `sources` key in build info was ordered alphabetically by path instead of by source unit ID, which broke compatibility with solc's JSON output and Hardhat's build format. Tools that rely on the index of a source path in `sources` matching its source unit ID (e.g., for source map resolution) would fail.

## Changes

1. **Custom serializer for `CompilerOutput.sources`** - Added a `sources_by_id` serde module that serializes the sources map ordered by source unit ID rather than by path
2. **Enable `preserve_order` for `serde_json`** - This ensures that when the output is converted to `serde_json::Value` (in `RawBuildInfo::new`), the insertion order is preserved
3. **Regression test** - Added a test that verifies sources in the output JSON are ordered by their source unit ID

## Before

```json
{"output": {"sources": {"a_first.sol": {"id": 2}, "m_middle.sol": {"id": 1}, "z_last.sol": {"id": 0}}}}
```

## After

```json
{"output": {"sources": {"z_last.sol": {"id": 0}, "m_middle.sol": {"id": 1}, "a_first.sol": {"id": 2}}}}
```

Note: The input sources remain alphabetically ordered as that's the input to the compiler. The output sources are now ordered by ID to match solc's output.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,7 @@ rayon = "1.11"
 regex = "1.11"
 semver = { version = "1.0", features = ["serde"] }
 serde = { version = "1", features = ["derive", "rc"] }
-serde_json = "1.0"
+serde_json = { version = "1.0", features = ["preserve_order"] }
 similar-asserts = "1"
 solar = { package = "solar-compiler", version = "=0.1.8", default-features = false }
 svm = { package = "svm-rs", version = "0.5", default-features = false }
--- a/crates/artifacts/solc/src/lib.rs
+++ b/crates/artifacts/solc/src/lib.rs
@@ -16,61 +16,6 @@ use std::{
     str::FromStr,
 };
 
-mod sources_by_id {
-    use crate::SourceFile;
-    use serde::{
-        de::{MapAccess, Visitor},
-        ser::SerializeMap,
-        Deserializer, Serializer,
-    };
-    use std::{collections::BTreeMap, fmt, path::PathBuf};
-
-    pub fn serialize<S>(
-        sources: &BTreeMap<PathBuf, SourceFile>,
-        serializer: S,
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -53,7 +53,7 @@ rayon = "1.11"
 regex = "1.11"
 semver = { version = "1.0", features = ["serde"] }
 serde = { version = "1", features = ["derive", "rc"] }
-serde_json = "1.0"
+serde_json = { version = "1.0", features = ["preserve_order"] }
```
zerosnacks: This introduces some level of overhead but should be relatively small. I am unsure however if this causes any unexpected side effects.

